PR #1357, opened 2015-02-04
Fix geom computation

gutard:
Triggers are not correct with line topologies with offset. They confuse mandatory points on line topologies with point topologies.


leplatrem:
Well done ! :clap: 

I suppose you had a hard time for this: you deserve a good hug next time I'll meet you in Toulouse :)

camillemonchicourt:
Nice :+1:

PR #1359, opened 2015-02-10
Fix snapping when a path has very close points

gutard:
If path A has very close points A1 and A2 (less than snapping distance). If path B has a point B1 at the same location as A1. The current algorithm may choose to snap B1 to A2. If A1 was path end, A will be splitted at A2. The new path A' between A1 and A2 will be very short and A1 may snap to B1 ie to A1. So the new path will get a null length and become an invalid geometry -> crash! This fix ensures that the snapping points always the closest.

### DIFF
--- a/geotrek/core/sql/50_troncons_split.sql
+++ b/geotrek/core/sql/50_troncons_split.sql
@@ -5,6 +5,7 @@ DECLARE
     linestart geometry;
     lineend geometry;
     other geometry;
+    closest geometry;
     result geometry;
     newline geometry[];
     d float8;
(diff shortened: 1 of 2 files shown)


camillemonchicourt:
Well done !

leplatrem:
Oh man it looks like you're only struggling with a lot of corner cases :) I hope you have fun on cool new features though :)

I may become a bit oppressive about this, but I strongly advise you (*for your mental health too* :smile:) to reproduce the bug in the tests suites (snapping was light I admit https://github.com/makinacorpus/Geotrek/blob/master/geotrek/core/tests/test_models.py#L142)

leplatrem:
As for the travis errors... welcome with the new version of flake8 :) No more named lambdas! https://github.com/mozilla-services/readinglist/pull/97/files

PR #1363, opened 2015-02-16
Use ubuntu packages and python eggs cache to speed up travis

gutard:


leplatrem:
You can try sudo: false in .yml also.. It might help to speed up travis

gutard:
We can't set sudo: false since we use sudo.

Actually, the problem is not to speed up travis but just make it work!
Most builds ends up with "Connection reset by peer" when downloading pip packages. 
See https://github.com/travis-ci/travis-ci/issues/2389

PR #1415, opened 2015-03-25
add interpreter

kiorky:


leplatrem:
I believe you can get rid of ``bpython`` in buildout-dev then

kiorky:
no, this is meaned to have a custom interpreter with the appropriate sys.path setted up, to execute arbitrary custom scripts (which will and wont have to be in sys.path at exec time).
another option would have been to use django-extensions/runscript, but it is just a sledgehammer for the small scripts we have to exec here.

leplatrem:
yes, that was the purpose of bpython in buildout-dev :)

PR #1432, opened 2015-04-20
Mention FR and EN languages to be able to load INITIAL DATAS

camillemonchicourt:


gutard:
Thanks. But the right process is to merge pull requests into develop, not into master. Please open next pull request on develop branch instead of master.

camillemonchicourt:
OK right !
I'll do that.

PR #1442, opened 2015-04-30
Trek to touristic content association distance depending on trek practice

gutard:


camillemonchicourt:
This feature has been included in version 2.0.0, right ?
I saw that field `distance` has been added to table `rando.o_b_pratique` but how to use it ? 
The value is the buffer in meter ?

gutard:
Never mind

PR #1458, opened 2015-06-23
Integration MapEntity Weasyprint

Babawah:


geotrek/core/views.py
```diff
@@ -117,13 +118,14 @@ def get_context_data(self, *args, **kwargs):
         return context
 
 
-class PathDocument(MapEntityDocument):
-    model = Path
+if not app_settings['MAPENTITY_WEASYPRINT']:
+    class PathDocument(MapEntityDocument):
```
leplatrem: This is wrong, defining classes upon settings is not a good idea!

geotrek/trekking/tests/test_views.py
```diff
@@ -345,24 +346,25 @@ def test_trek_export_poi_list_limit(self, mocked_prepare, mocked_attributes):
 
 class TrekCustomPublicViewTests(TrekkingManagerTest):
 
-    @mock.patch('mapentity.models.MapEntityMixin.get_attributes_html')
-    def test_overriden_document(self, get_attributes_html):
-        trek = TrekFactory.create()
+    if not app_settings['MAPENTITY_WEASYPRINT']:
```
leplatrem: I am sad, this was merged  without any test. The only modification was this one and this is meant to disable existing ones...

In MapEntity the only test consists in checking that the whole thing returns a 200 status code. That means that not a single "if" of those hundreds of lines is tested.

I didn't want at first to come here and teach a lesson, but I'm sorry I couldn't resist:
If you think that you save time (and money) when bypassing tests, you are wrong. You loose quality assurance, stability, flexibility for refactoring, confidence when upgrading dependencies, and even prestigious image we had. And of course users will start to be angry when regressions will appear, so that is the best way to be even more stressed, with tedious development and thus deliver late.

PR #1461, opened 2015-06-25
Transform POI coords in GPX in projection 4326

camillemonchicourt:
Should fix - https://github.com/makinacorpus/Geotrek/issues/1351

### DIFF
--- a/geotrek/trekking/serializers.py
+++ b/geotrek/trekking/serializers.py
@@ -29,9 +29,10 @@ class TrekGPXSerializer(GPXSerializer):
     def end_object(self, trek):
         super(TrekGPXSerializer, self).end_object(trek)
         for poi in trek.pois.all():
-            wpt = gpxpy.gpx.GPXWaypoint(latitude=poi.geom.y,
-                                        longitude=poi.geom.x,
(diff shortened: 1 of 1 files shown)


geotrek/trekking/serializers.py
```diff
@@ -29,9 +29,10 @@ class TrekGPXSerializer(GPXSerializer):
     def end_object(self, trek):
         super(TrekGPXSerializer, self).end_object(trek)
         for poi in trek.pois.all():
-            wpt = gpxpy.gpx.GPXWaypoint(latitude=poi.geom.y,
-                                        longitude=poi.geom.x,
-                                        elevation=poi.geom.z)
+            geom3d = poi.geom.transform(4326, clone=True) # GPX uses WGS84
```
leplatrem: Add two spaces before ``#``

leplatrem:
It would be useful to add some tests that make sure it actually fixes the problem :)

leplatrem:
The fix was confirmed to be successful on the mailing-list :)

gutard:
Superseeded by PR #1471 that adds tests and correctly handle elevation of waypoints. I close this one.

PR #1470, opened 2015-07-29
Save geom only if updated

Anaethelion:
Add new logo
Fix minors bugs
Add logs output on fail for travis
Fix output for imports hml/txt for command debug purposes


geotrek/common/parsers.py
```diff
@@ -292,7 +292,8 @@ def report(self):
             'nb_unmodified': self.nb_unmodified,
             'warnings': self.warnings,
         }
-        return render_to_string('common/parser_report.html', context)
+        print(output_format)
```
leplatrem: I recommend you to use logging :)

geotrek/common/parsers.py
```diff
@@ -292,7 +292,8 @@ def report(self):
...
+        return render_to_string('common/parser_report.{output_format}'.format(output_format=output_format), context)
```
leplatrem: In this case, using ``%s`` is acceptable

geotrek/common/parsers.py
```diff
@@ -282,7 +282,7 @@ def parse_row(self, row):
         if self.progress_cb:
             self.progress_cb(float(self.line) / self.nb, self.line, self.eid_val)
 
-    def report(self):
+    def report(self, output_format='txt'):
```
leplatrem: The txt format is not tested (see decreasing coverage)

geotrek/trekking/models.py
```diff
@@ -269,6 +269,20 @@ def distance(self, to_cls):
     def is_public(self):
         return self.any_published or (self.parent and self.parent.any_published)
 
+    def save(self, *args, **kwargs):
+        if self.pk is not None and kwargs.get('update_fields', None) is None:
+            field_names = set()
+            for field in self._meta.concrete_fields:
+                if not field.primary_key and not hasattr(field, 'through'):
+                    field_names.add(field.attname)
```
leplatrem: The update fields logics implemented here is not tested

geotrek/common/parsers.py
Anaethelion: This, is a mistake, good catch !

geotrek/common/parsers.py
Anaethelion: Will do.

geotrek/trekking/models.py
Anaethelion: It is the same logic as django's but i'll see what i can do.

geotrek/trekking/tests/test_views.py
```diff
@@ -1064,3 +1064,29 @@ def test_can_publish(self):
         self.assertContains(response, 'Published')
         response = self.client.get('/trek/edit/%u/' % self.trek.pk)
         self.assertContains(response, 'Published')
+
+
+class TrekUpdateGeomTest(TestCase):
```
leplatrem: This is a detail, but this test belongs more to test_models :)

PR #1488, opened 2015-09-16
fixes #1478 Handle multiple reservation systems

### LINKED ISSUE #1478: Handle multiple reservation systems
- [ ] Add a new reservation_system field
- [ ] Add new field in API
- [ ] Allow to see/edit this field
- [ ] Ascending compat
- [ ] Documentation
- [ ] Test

submarcos:


geotrek/tourism/admin.py
```diff
@@ -64,3 +64,10 @@ class TouristicEventTypeAdmin(TranslationAdmin):
 
 if settings.TOURISM_ENABLED:
     admin.site.register(tourism_models.TouristicEventType, TouristicEventTypeAdmin)
+
+
+class ReservationSystemAdmin(admin.ModelAdmin):
+    list_display = ('name',)
+    search_fields = ('name',)
+
+admin.site.register(tourism_models.ReservationSystem, ReservationSystemAdmin)
```
gutard: Should register only if settings.TOURISM_ENABLED is set

gutard:
Maybe in API (/api/fr/touristiccontents.geojson), the reservation system should be the name instead of the id.

PR #1490, opened 2015-09-17
Check for size change instead of name for attachments already downloaded

Anaethelion:


geotrek/common/parsers.py
```diff
@@ -466,6 +469,26 @@ def filter_attachments(self, src, val):
...
+                ftp.login(user=parsed_url.username, passwd=parsed_url.password)
+                ftp.cwd(directory)
+                size = ftp.size(parsed_url.path.split('/')[-1:][0])
+                return size == attachment.attachment_file.size
+
+            if parsed_url.scheme == 'http' or parsed_url.scheme == 'https':
+                http = urllib2.urlopen(url)
+                size = http.headers.getheader('content-length')
+                return size == attachment.attachment_file.size
+        except:
```
leplatrem: Catch-all are bad

geotrek/common/parsers.py
```diff
@@ -466,6 +469,26 @@ def filter_attachments(self, src, val):
...
+        try:
+            parsed_url = urlparse(url)
+            if parsed_url.scheme == 'ftp':
+                directory = dirname(parsed_url.path)
+
+                ftp = FTP(parsed_url.hostname)
+                ftp.login(user=parsed_url.username, passwd=parsed_url.password)
+                ftp.cwd(directory)
+                size = ftp.size(parsed_url.path.split('/')[-1:][0])
+                return size == attachment.attachment_file.size
```
leplatrem: This will return true is size is the same. The fonction name is "has changed" though

leplatrem:
Definitely would deserve tests.

geotrek/common/parsers.py
gutard: Yes, I already catched and fixed that.

leplatrem: Grea! sorry then :( I didn't dedicate much time TBH

PR #1491, opened 2015-09-17
Backport to V1.x changes made to attachments size check on import.

Anaethelion:


camillemonchicourt:
Thanks, does it means that another V1.X will be released ?

camillemonchicourt:
I have to update my Geotrek-admin 0.35.
Not sure if I should use v1.0.0 or include these commits and use https://github.com/makinacorpus/Geotrek/tree/v1.x ZIP ?

PR #1496, opened 2015-09-23
fixes #1479 children treks ordering

### LINKED ISSUE #1479: Children treks ordering
And allow one child in multiple parent treks

- [ ] Replace parent field by children one
- [ ] New widget
- [ ] Adapt exports, API
- [ ] Test
- [ ] Doc

submarcos:


geotrek/tourism/models.py
```diff
@@ -281,6 +281,7 @@ def __unicode__(self):
         return self.name
 
     class Meta:
+        #db_table = 't_b_systeme_reservation'
```
gutard: Should be moved in its own commit (uncommented, with the corresponding south migration)

geotrek/trekking/forms.py
```diff
@@ -64,6 +64,12 @@ class Meta(CommonForm.Meta):
 
 
 class TrekForm(BaseTrekForm):
+    children_trek = forms.ModelMultipleChoiceField(label=_(u"Children"),
+                                                   queryset=Trek.objects.all(), required=False,
```
gutard: Maybe we should filter by user structure or trek structure. I'll ask Camille.

PR #1499, opened 2015-09-25
fixes #1494 Condition infrastructure

### LINKED ISSUE #1494: Add CONDITION field in INFRASTRUCTURE table
So it would be available in SIGNAGE and INFRASTRUCTURE modules.
"Etat" in French.
It would be a list of values. 

So in DB it would add : 

- `etat` field in `gestion.a_t_amenagement` (Foreign key)
- `gestion.a_b_etat` table (id, structure, etat)

Also to be included in detail pages, detail and list exports and filters. 

PS : We would also need it in Geotrek-admin V1.X so need to discuss if it is possible to deploy in V1 and V2.

submarcos:


geotrek/infrastructure/models.py
```diff
@@ -53,6 +53,18 @@ def __unicode__(self):
         return self.label
 
 
+class InfrastructureState(StructureRelated):
+    label = models.CharField(verbose_name=_(u"Name"), db_column="etat", max_length=250)
+
+    class Meta:
+        verbose_name = _(u"Infrastucture State")
+        verbose_name_plural = _(u"Infrastucture States")
```
gutard: Typo

PR #1506, opened 2015-10-05
Parkings as services

Anaethelion:
Remove parking feature from treks and add it to services.
Add data migration.
Add cache for buildout deps.
Add fix to prevent error on service add if no path exists in database.


camillemonchicourt:
It means that we dont localize parkings anymore when we create a trek ?
And what about labels on Parking highlight ?

parking

Anaethelion:
Hello Camille,

This has been made as per request from Cevennes's requirements, i thought that was agreed beforehand.
If need be, i can make it so the two functionalities work along each others.

The goal was to have the parkings in geotrek-rando in a nice popup for visitors.
Let met know what you thinks !

camillemonchicourt:
OK I'll talk about it with them today to see the consequences.

camillemonchicourt:
OK, I talked with PNC and they also dont understand. 
It is very important to associate ONE parking with each trek. It cant work automatically with a buffer. 
V1 system was perfect.

PR #1508, opened 2015-10-07
fixes #1400 Clarify 3d 2d lengths

### LINKED ISSUE #1400: Clarify 3D and 2D lengths
It is a good point that both lengths are now displayed in detail pages.

On lists, this is the 3D length that is displayed, good point too but it would be good to mention it in label (3D length).

In list exports, it would be great to have both length too. As actually it seems like there is only the 3D length in exports.

submarcos:


camillemonchicourt:
It seems OK in CSV exports but not in SHP exports that don't use good labels.
They display "Length_2D" and "Length". In English it should be "2D_length" and "3D_length" as done in this PR.

Also all labels in SHP are in English while they are translated in CSV. Is that normal ?
Thanks.

PR #1509, opened 2015-10-08
Add altimetry svg limits to json to be used in geotrek - rando

Anaethelion:


leplatrem:
Nice to see that you wrote some tests for the new code! GG!

PR #1518, opened 2015-10-13
Filter touristic events geojson generation based on export date.

Anaethelion:


geotrek/tourism/filters.py
```diff
@@ -32,3 +32,11 @@ class Meta(StructureRelatedFilterSet.Meta):
...
         fields = StructureRelatedFilterSet.Meta.fields + [
             'published', 'type', 'themes', 'after', 'before', 'approved']
+
+
+class TouristicEventApiFilterSet(django_filters.FilterSet):
+    ends_after = django_filters.DateFilter('end_date', lookup_type='gte')
+
+    class Meta:
+        model = TouristicEvent
+        fields = ['end_date', ]
```
leplatrem: This is easy to test, yet I see no test file modified in the PR :)

Anaethelion: Ain't aa64837 what you are referring to ? :)
If not i would gladly add one !

leplatrem: Not exactly, I would suggest a test where you create a few records and check that when you call the view with the filter you don't obtain every of them.
Think the tests as a specification of what you implemented.
Like if you removed or change this new code, some test should fail somewhere.
It really helps for long term maintenance!

Anaethelion: Obviously you haven't seen my edit.
For the record, i mistakenly commited the test to another branch on my machine so i have to tidy things up before adding to this PR.

leplatrem: Okay! Sorry then!

Great if you did something similar! 

Enjoy ;)

Anaethelion: No harm done, besides you made me realize i made a tremendous mistake with my branches, ah too many features at once is not a good idea. ;)

PR #1522, opened 2015-10-21
sync_rando view and user feedback

submarcos:


geotrek/trekking/templates/trekking/sync_rando.html
```diff
@@ -0,0 +1,139 @@
...
+	<link rel="stylesheet" href="{% static 'trekking/css/sync_trek.css' %}" />
+{% endblock extrahead %}
+
+{% block toolbar %}
+{% endblock toolbar %}
+
+{% block mainpanel %}
+
+	<div id="sync-div" class="span3 offset2">
+	    <h3>{% trans "Trekking sync" %}</h3>
```
gutard: Web site and mobile app synchronization (not only trekking)

geotrek/trekking/templates/trekking/sync_rando.html
```diff
@@ -0,0 +1,139 @@
...
+		            <button type="button" class="btn btn-default" data-dismiss="modal">
+		                {% trans "Cancel" %}
+		            </button>
+		            <a href="#" id="btn-confirm" class="btn btn-success success">
+		                {% trans "Sync" %}
+		            </a>
+		        </div>
+	    	</div>
+		</div>
+	</div>
```
gutard: The confirmation modal is not required. Before there was no confirmation after click in menu, now there are 2 confirmations. 1 is sufficient.

gutard:
We should show if a sync is pending (not only in progress). It seems possible to inspect task queue:
http://stackoverflow.com/questions/5544629/retrieve-list-of-tasks-in-a-queue-in-celery

PR #1530, opened 2015-11-02
Fix django debug toolbar

submarcos:
fix settings and urls to work with django debug toolbar in dev environment


gutard:
Does it works with prod settings but DEBUG=True ?

submarcos:
No it doesn't, debug_toolbar app and middleware are declared in settings.dev,
so with prod and DEBUG=True, nothing happend (However, django-debug-toolbar must be installed because urls.py import debug_toolbar in case of DEBUG=True).

It was originaly declared in settings.dev, but I think it can be moved in settings.prod or settings.default. It is disabled with Debug=False in production

PR #1537, opened 2015-11-23
Switch map backend

submarcos:


geotrek/settings/default.py
```diff
@@ -124,6 +124,9 @@
     MAP_STYLES.setdefault('detail', {})['color'] = _others_color
     MAP_STYLES.setdefault('others', {})['color'] = _others_color
 
+FORCE_OSMBASELAYER_BOUNDINGBOX = []
```
gutard: I think we should provide a more generic setting (not bound to OSM only).
Currently we have a settings for layers, eg:
LEAFLET_CONFIG['TILES'] = [..., ('OSM', 'http://{s}.tile.osm.org/{z}/{x}/{y}.png', '(c) OpenStreetMap Contributors'), ...]
So our setting could be:
FORCE_LAYERS = [..., ('OSM', [(lat1, lng1), (lat2, lng2), (lat3, lng3)], ...]
to allow to force multiple polygons to any defined layer by referencing its name.

geotrek/settings/default.py
```diff
@@ -124,6 +124,9 @@
...
+# [(lat1, lng1), (lat2, lng2), (lat3, lng3)], (lat4, lng4)]
```
gutard: Extraneous ] in the middle ?

geotrek/templates/mapentity/base.html
```diff
@@ -0,0 +1,57 @@
...
+	            		
+	            		if(polygon.getBounds().contains(map.getCenter())){
+			    			// switch std map
+			    			map.removeLayer(osm);
+			    			//console.log('in');
+			    		}
+			    		else {
+			    			// switch OSM map
+						    map.addLayer(osm);
+						    //console.log(map.getCenter() + ' not in (' +polygon.getBounds()._northEast+' - '+polygon.getBounds()._southWest);
```
gutard: Debug/commented code should not go to master

PR #1548, opened 2015-12-01
Add a command to clean orphaned attached media

gutard:


camillemonchicourt:
Great !

gutard:
But completely bugged and likely to remove all your media file :(
So I reverted it.

camillemonchicourt:
Ahahahaha, so better to revert indeed.

PR #1565, opened 2016-01-18
fix template tag duration

submarcos:


gutard:
TREK_DAY_DURATION definition should be removed from geotrek/settings/base.py

PR #1575, opened 2016-02-03
Fix structure in objects with user structure instead of default structure

submarcos:


gutard:
Already merged manually with a fast forward not detected by github

PR #1577, opened 2016-02-04
Set Weasyprint as default PDF engine

gutard:


camillemonchicourt:
Done in 2.10.4

gutard:
This PR totally remove ODT templates but we want to keep them for private exports. So I close it without merging.

camillemonchicourt:
Yes indeed, as discussed here - https://github.com/makinacorpus/Geotrek/issues/1589 and done in 2.10.4. Thanks.

PR #1585, opened 2016-02-29
delete tilecache references, fix default leaflet tiles

submarcos:


camillemonchicourt:
OK so it would close this one - https://github.com/makinacorpus/Geotrek/issues/1184 ?
It is a good thing as most structures are using IGN Geoportail API. 
I guess that advanced configuration becomes manadatory to set the map tiles - http://geotrek.readthedocs.org/en/master/advanced-configuration.html#using-google-maps-projection-in-the-web-interface
Then it would make sense to include this setting in ``setting.ini`` and to remove the WMS part (https://github.com/makinacorpus/Geotrek/blob/master/conf/settings.ini.sample#L37-L44) that will not work anymore with this commit. Good thing too but to remove in this case. 
Thanks.

gutard:
> I guess that advanced configuration becomes mandatory to set the map tiles

No. The default config now comes with DB in Lambert-93 projection and leaflet with OSM tiles in web mercator projection.

Advanced configuration is needed only to set up IGN tiles

PR #1607, opened 2016-07-13
Mention filtering by portal in doc

camillemonchicourt:
New option available in sync_rando command to filter by portal


camillemonchicourt:
Duplicate

PR #1629, opened 2017-01-25
Update github URL and clarify installation

camillemonchicourt:


camillemonchicourt:
Sorry for the dedicated branch, it can be deleted after merging. 
I am not very good with updating my fork in order to do some PR...

Issue #952, opened 2014-02-09
Command to load a point layer as linear reference records


leplatrem:
This Django command would take a layer with point geometries and a model name, and create relevant topologies according the path network.

For example: 

```
bin/django loadtopo geotrek.trekking.POI fountains.shp
bin/django loadtopo geotrek.infrastruture.Signage pancartes.shp
```

For points, it's very easy, since the method `Topology.deserialize()` already converts lat/lng to topologies.

Use `from django.db.models.loading import get_model`
See https://docs.djangoproject.com/en/1.6/howto/custom-management-commands/

camillemonchicourt:
I created a SHP layer to test it with only one POI. I uploaded it on the server and launched ``bin/django loadpoi ../mypois.shp``.

It worked well after I transformed the layer into 4326 but it didnt included the name and type even if I had the fields `name` and `type` in the layer :

![poi-layer](https://cloud.githubusercontent.com/assets/4418840/16532832/d8e882ce-3fd4-11e6-8257-c30a0adfe14c.jpg)

Would be great to fix this, to allow it for other modules with points (signages, infrastructures, interventions...) and to include it in the new IMPORT module as discussed here - https://github.com/makinacorpus/Geotrek/issues/1595

TheoLechemia:
I have the same problem as Camille using the command `loadpoi` . The object is created in Geotrek but without his attributes.
However my goal was to import 'Infrastructures'...
I adapt the 'loadpoi' command to import infrastructures (aménagement), and it works, attributes are also imported.
Should I do a pull request to add this new command in the project ?

camillemonchicourt:
Yes sure, very interested with it !!!!

camillemonchicourt:
OK so there has been some improvement with a new command in 2.12.0 (https://github.com/GeotrekCE/Geotrek-admin/releases/tag/2.12.0) to load SIGNAGES from a SHP (?) into Geotrek database. 
There is also a pending PR to load infrastructures (not only signages) but it is not finished and has now some conflicts : https://github.com/GeotrekCE/Geotrek-admin/pull/1630

And now @CynthiaBorotPNV from PnVanoise did an interesting Python to insert her PostGIS datas from a previous database into her Geotrek database : http://geotrek.ecrins-parcnational.fr/ressources/technique/2017-10-13-impFromAigleDB.py

To be clarified in upcoming versions.

leplatrem:
[...]

PR #1630, opened 2017-01-26
add django command to load infrastructure from shp

TheoLechemia:
I just adapt the `loadpoi` django command to load 'Infrastuctures' (aménagement) from a shapefile

The shapefile must have this three fields:
-name
-type
-etat

### DIFF
--- a/geotrek/infrastructure/management/commands/loadinfrastructures.py
+++ b/geotrek/infrastructure/management/commands/loadinfrastructures.py
@@ -1,0 +1,76 @@
+import os.path
+
+from django.core.management.base import BaseCommand, CommandError
+from django.contrib.gis.geos import GEOSGeometry
+
+from geotrek.core.helpers import TopologyHelper
+
+from geotrek.infrastructure.models import Infrastructure, InfrastructureType, InfrastructureCondition
+
+
+class Command(BaseCommand):
+    args = '<point_layer>'
+    help = 'Load a layer with point geometries in a model\n'
+    can_import_settings = True
+    field_name = 'name'
+    field_infrastructure_type = 'type'
+    field_condition_type = 'etat'
+
+    def handle(self, *args, **options):
+
+        try:
+            from osgeo import gdal, ogr, osr  # NOQA
+        except ImportError:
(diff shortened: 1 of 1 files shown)


camillemonchicourt:
Great but it still mention poitype. Normal ?

TheoLechemia:
Ahhh sorry.... I forgot to change some variable names...

camillemonchicourt:
It looks like some CODESTYLE and INTEGRATION tests failed ?

TheoLechemia:
I'll check the codestyle since I have my serveur back...
For integration, I have no idea why ...
Someone has any idea ?

gutard:
> For integration, I have no idea why ...
> Someone has any idea ?

You can ignore. It is allowed to fail in travis

camillemonchicourt:
Related issue : https://github.com/GeotrekCE/Geotrek-admin/issues/952

camillemonchicourt:
@submarcos just made the similar command to load signages : https://github.com/GeotrekCE/Geotrek-admin/commit/4da515d

Signages are a category of infrastructures, in the same table, so I dont know if these 2 commands are complementary or partially similar ?

Also here are a few ideas :

- Add a test with a sample SHP to verify that it is well imported into database
- Command fails if some fields are empty ?
- Check projection

camillemonchicourt:
@submarcos just made the similar command to load signages : https://github.com/GeotrekCE/Geotrek-admin/commit/4da515d

Signages are a category of infrastructures, in the same table, so I dont know if these 2 commands are complementary or partially similar ?

Also here are a few ideas :

- Add a test with a sample SHP to verify that it is well imported into database
- Command fails if some fields are empty ?
- Check projection

TheoLechemia:
If I understand well, this command is a bit more powerfull !
- You can pass the name of the related field names (type, conditions...) in the command
- It can read all formats supported by OGR (geojson, shp..)
- it reprojects the file with the good projection

They only differente I see, is the call to the ORM 
for signages `Signage.objects.create`
for infrastructures: `Infrastructure.objects.create`
but for the rest, it can be factorize in a function for example

Issue #1620, opened 2016-11-14
Ponctual events and layers intersection


camillemonchicourt:
We created a signage that is far from paths. 
The signage is in the middle of PRUNIERES municipality but in indicated in SAINT-APPOLINAIRE municipality : 

![point-municipality](https://cloud.githubusercontent.com/assets/4418840/20264463/fb5f702c-aa6b-11e6-99d5-9ab85c123eea.jpg)

I guess that it is because the PK where it is rattached on path 2833 is in SAINT-APPOLINAIRE municipality as we can see in the image. But it should use the offset (that is big in this case) when intersecting with polygon layers. 

Or use the geometry of the event ?

camillemonchicourt:
Fixed in 2.11.5.

PR #1631, opened 2017-01-31
Fix infrastructure zoning

submarcos:
- fix intersection between topologies and cities

- enhance zoning admin


camillemonchicourt:
Thanks, is it fixing https://github.com/GeotrekCE/Geotrek-admin/issues/1620 ?

submarcos:
Yes it is. In some cases there are no city at all. But I need to test anymore before merging

camillemonchicourt:
OK great, thanks !

PR #1651, opened 2017-06-16
Fix sql and pre/post migrate

submarcos:
- Fix sql path split functions
- Fix pre and post migrate according django 1.8 recommandations


submarcos:
need to revert and translate again before merging

PR #1720, opened 2018-03-28
Remove landplanner basemap config

wboykinm:
With apologies; I have to discontinue this map layer. It has become responsible for numerous charges on my Mapbox account originating from various sources.


camillemonchicourt:
Yes it has to be removed from examples. Thanks.

PR #1762, opened 2018-07-16
Perf altimetry

LePetitTim:
Don't update 3d geom if the geometry is the same as before

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -234,6 +234,7 @@ CHANGELOG
 
 - Add a settings allowing to remove certain items from the left menu
 - Serve attachment with 'Topoguide' type as public PDF
+- Avoid calculate altimetry when the geometry didn't change.
 
 **Bug fixes**
 
@@ -319,7 +320,7 @@ CHANGELOG
 2.27.6 (2019-06-04)
 -------------------
 
-**Bug fixes**
+**Minor changes**
 
 - Fix mobile API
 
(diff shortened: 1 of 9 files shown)


camillemonchicourt:
Great ! Thanks.

LePetitTim:
I am trying a solution to make less calcul for altimetry for example : 
- Do not calculate altimetry for all the path but only for the part modified.
- Same for event, topologies.
For elevation only 3D linestrings max and min, the calculation is not that hard but for slope, positive gain ... it is. We need to calculate for all the linestring again. If you have any suggestion it would be great.

LePetitTim:
Trek's geometries were computed a second time even if informations were in the base in Path (e_t_troncon), now we just get the 3D part. Now changing a trek take less than 10 secs (10mins before).

camillemonchicourt:
Wahooo

LePetitTim:
With this modification : 30s , without 1min30 (on 1 path with 2 trails)
With : 1 s, 7 s (on 1 trek)

camillemonchicourt:
Thanks, seems like a great improvement !

LePetitTim:
I will try to improve the split part. Right now, we compute the altimetry for the one split but actually, a path splitted by another one do not move, this is just split so the altimetry should be compute with the previous altimetry (3D line) and not recomputed.

LePetitTim:
Now, during the split, if we split a path really long, it doesn't recompute the path splitted.
For example : a path of 500km split with a path of 20 km take 3 minutes without
and 5 s with

camillemonchicourt:
Still so good to hear this kind of news !

camillemonchicourt:
Thanks, replaced by https://github.com/GeotrekCE/Geotrek-admin/pull/2195, included in version 2.33.

PR #1768, opened 2018-08-13
Add dailymotion in URLs attachments

LePetitTim:


gutard:
Squashed and merged manually

PR #1770, opened 2018-08-20
Add filter comfort and stake

LePetitTim:
Add the filter of confort and stake in Path filters (on the left). We keep in the right filters of topologies (treks, intervention ...).

Asked on this issue : #1668


camillemonchicourt:
Thanks

PR #1772, opened 2018-08-22
Add structures in filter 

LePetitTim:
Add the structure to NoneRelated Object name. Allowing to see difference between 2 elements with same name but with different structure.
image

Ref : #1766


camillemonchicourt:
OK so the change is in filters ? As the screenshot shows a detail page. So I am not sure to understand what this feature is about.

LePetitTim:
Avant : 
image
Apres : 
image

camillemonchicourt:
Perfect, thanks a lot.

PR #1817, opened 2018-12-14
WIP: Add docker section in installation doc

numahell:
Update documentation with docker installation

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -17,9 +17,117 @@ A first estimation on system resources is :
 * 4 Go RAM
 * 20 Go disk space
 
+Docker Installation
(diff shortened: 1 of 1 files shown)


camillemonchicourt:
Thanks

camillemonchicourt:
That's a great thing if Docker installation is working now. 
But in documentation I would leave manual installation first as for most of actual Geotrek users it is the easiest way to install Geotrek.

submarcos:
It's already in docs install, but before merging All documentation must be completed and refectored

camillemonchicourt:
OK thanks for the details.
Very interesting.

PR #1826, opened 2018-12-31
Add tests 3

LePetitTim:


LePetitTim:
I am trying again to add some tests. More I am doing this more i found some code which are never use. It would be great to check if this is still usefull.

PR #1843, opened 2019-01-17
Do not display fragment on same model

numahell:
For infrastructures and signages


gutard:
> There are more than this two. I think it's on all models.

Yes, but an object can be linked to another object of the same type. For example a touristic content is linked to other touristic contents within 500 meters.

Maybe we should just hide the <h1> if there is no content inside.

PR #1951, opened 2019-05-14
Remove POI public PDF

gutard:


camillemonchicourt:
It means that POI are now removed from Trek public PDF for everyone ?
POI can take a few pages but they are crucial for many of us on public PDF.

gutard:
Non ça supprime juste le bouton "Impr." sur la fiche détail du POI dans Geotrek-admin qui de toute façon crashe.

camillemonchicourt:
Ah OK ! Merci, bien vu.

PR #1981, opened 2019-06-06
Fix TouristicContent kept with filters + remove additional layer on the same module

LePetitTim:


LePetitTim:
ToursticContent layer on top left are still charged even when we try to filter !!

PR #2051, opened 2019-08-27
WIP : Add a command to import dives

gutard:


gutard:
@LePetitTim will add some tests

PR #2065, opened 2019-09-23
Add diving rando

LePetitTim:


geotrek/diving/serializers.py
```diff
@@ -43,6 +52,19 @@ class DiveSerializer(PicturesSerializerMixin, PublishableSerializerMixin,
     source = RecordSourceSerializer(many=True)
     portal = TargetPortalSerializer(many=True)
     category = rest_serializers.SerializerMethodField()
+    dives = CloseDiveSerializer(many=True, source='published_dives')
+    treks = CloseTrekSerializer(many=True, source='published_treks')
```
numahell: We'll may have to add POIs, as in geotrek/tourism/serializers.py

geotrek/diving/templates/diving/dive_detail_attributes.html
```diff
@@ -40,10 +40,6 @@ <h3>{% trans "Attributes" %}</h3>
             <th>{{ object|verbose:"facilities" }}</th>
             <td>{{ object.facilities|safe }}</td>
         </tr>
-        <tr>
```
numahell: Why removing this part ?

numahell: Ok it was in double, have not noticed sorry.

PR #2099, opened 2019-11-20
Move to Python 3 (most work by @LePetitTim)

gutard:


camillemonchicourt:
Thanks for this big work

PR #2127, opened 2020-01-16
Virtualenv install

gutard:
Add old install.sh to docker-integration branch and tweak it to make it work


gutard:
Superseded by debian packaging

PR #2140, opened 2020-02-11
Remove all db column and db table

LePetitTim:


camillemonchicourt:
International ! :-)
Then will have to be updated with "Remove NoDeleteMixIn" to remove the references to "deleted" field and "deleted = FALSE" conditions.

PR #2224, opened 2020-04-27
Test package

gutard:


gutard:
It seems not possible to get a VM with Ubuntu 18.04 in CircleCI. And docker images do not work with systemd.

PR #2254, opened 2020-05-18
Feedback suricate

numahell:
PR for feedback suricate (same as #2185 but with correct branch)


LePetitTim:
https://github.com/GeotrekCE/Geotrek-admin/pull/2257

Issue #2294, opened 2020-06-30
Migration <= 2.32 to packaged version 2.33 error `No module named 'geotrek.settings.prod'`


numahell:
Distro : Ubuntu 18.04
Migration from 2.32.11 to .2.33 with migration script

numahell:
Actually it is a regex error with sed in `migrate.sh`

gutard:
To be exact, this problem occur only if the file custom.py is modified on Windows (with CR+LF ends of line). Actually the regex is correct. This is the ends of line wich are not. If we modify the regexp, the purpose is only to tolerate the error from the user that do not configure its text editor correctly.

PR #2295, opened 2020-06-30
Fix migrate sh sed error

numahell:
Concerns #2294


camillemonchicourt:
Thanks for including this file in GIT repository. Would be good to also include the install.sh script into this directory

camillemonchicourt:
I also suggest to include a small content at the begining of the migrate.sh script to mention that it is used to migrate from a version before 2.33 to 2.33.

camillemonchicourt:
Thanks

PR #2326, opened 2020-09-15
add network alias to get screamshotter and convertit working in docke…

submarcos:
Revert changes in docker-compose to keep screamshotter and convertit working in development mode

to work, you need to define DOMAIN_NAME in dev mode (see .env.dist and your .env file)

for example, geotrek.local, and set your /etc/hosts to target 127.0.0.1 

Problem is that screamshotter and convertit can't resolve localhost or 127.0.0.1 in local docker development mode, and this address came from your browser url. So browser url should be resolved by screamshotter and convertit containers in local and not public environmenent)

This problem can also happen in production if private DNS are used (resolve by using extra_hosts)

Capture d’écran de 2020-09-15 13-53-50

### DIFF
--- a/.env-dev.dist
+++ b/.env-dev.dist
@@ -1,5 +1,5 @@
 ENV=dev
-DOMAIN_NAME=geotrek.local
+SERVER_NAME=geotrek.local
 POSTGRES_HOST=postgres
 POSTGRES_PORT=5432
 POSTGRES_USER=geotrek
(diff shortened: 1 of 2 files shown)


gutard:
J'ai renommé en SERVER_NAME pour être cohérent avec la conf de prod et j'ai ajouté une valeur par défaut dans le docker-compose.yml

gutard (approved):

PR #2330, opened 2020-09-16
use official postgis docker image

submarcos:
Existing volumes initiated by using makinacorpus/postgis:11-2.5 work well with postgis/postgis:11-2.5


gutard:
Nickel ! Je merge.

PR #2398, opened 2020-11-12
Docs reorg index

numahell:
Here is a first iteration to reorganize documentation. This pull request:

- adds in docker compose dev sphinx service to generate doc locally
- reorganizes TOC to show sections in sidebar menu : see the result here https://geotrek.readthedocs.io/en/docs-reorg-index/ 

Comments are welcome about the sections (order, level…).


camillemonchicourt:
Yes good idea to have sections as we can be glad that our documentation has a lot of parts now :-)
And usually there are different kind of documentation users and usages : 
- Using application (final users)
- Installing application (sysadmins)
- Managing application (DBA and sysadmins)
- Developping application (Developpers)

So your sections organization and order suggestion is perfect IMHO.

Thanks.

camillemonchicourt (approved):
Thanks

numahell:
Thanks for the feedback !
Indeed I also think managing application could be a dedicated section, but I did not want to make too many changes in docs for this first step.

PR #2403, opened 2020-11-17
Load poi geojson

babastienne:


camillemonchicourt:
SHP import is replaced by GeoJSON ?
Good to move from SHP

gutard:
> SHP import is replaced by GeoJSON ?

Before : only SHP / Now : SHP, GeoJSON or whatsoever (any format supported by GDAL)

camillemonchicourt:
OK. Even better.
In some other projects we are replacing SHP by GeoPackage and sometimes GeoJSON

gutard (approved):

PR #2427, opened 2020-12-03
Update installation.rst

mviadere-openig:
change 
   sudo apt-get update  --allow-releaseinfo-change' 

to    sudo apt-get update  --allow-releaseinfo-change


gutard:
Thanks

PR #2433, opened 2020-12-05
Delete old attachments by default in parsers

gutard:


numahell (changes requested):
Dev part is ok for me, but changelog is missing.

PR #2528, opened 2021-02-03
Fix missing id attributes in API v2

### LINKED ISSUE #2527: API V2 - Manque les id des sources sur la route /source/

gutard:
Fix #2527


camillemonchicourt:
Merci

PR #2551, opened 2021-02-17
Draft : Refactorize to work with sensitivity

submarcos:
recent changes for api mobile and api v2 breaks sensitivity instance

May be we should to add specific tests for particular instances (sensitivity only, others apps disabled... etc)

We need to statuate for zoning, which is now required by all instances


LePetitTim:
POC:

Show it doesn't work without some apps

PR #2574, opened 2021-03-06
Avoid a db connection when requesting time from database

claudep:
Typically when running tests, this was querying the non-test database,
which might exist or not.


gutard (approved):
Actually I'm not sure weed need dbnow() at all because timestamp is set later by a trigger.

PR #2584, opened 2021-03-10
Add select/prefetch related to exports

gutard:
Add select/prefetch related to FormatList views (exports). For now I removed zoning from attributes as it triggers a request for each row. Maybe we could add an annotate to handle this.


LePetitTim:
IMHO, it will not be possible to merge this PR with modifications on django-geojson + custom columns @submarcos @Chatewgne

PR #2666, opened 2021-05-25
Feat: Add OpenAir export format to  Sensitivity module (#2372)

lpofredc:
As mentionned in #2372, there are expectations for the OpenAir file format in the context of free flight activities. This merge request adds the ability to export sensitive zonings in the OpenAir format, either individually or for all sensitive areas.

This new feature stands on PyOpenair lib, specially developped for this.


babastienne:
Hi. Thank you very much for your contribution !

Could you fix the codestyle of the code ? (you can follow the guidelines at https://geotrek.readthedocs.io/en/master/development.html#conventions to match our linting rules)

Also, because this is a specific feature that use an external library, we try to ensure that this will not introduce regressions in the future. The library PyOpenair does not seems to be tested at all and should be before we can validate the PR. Because if there is a regression we should be able to contribute to the project (and unit tests helps to understand use cases).

If a new standard appears in the future (for example in the WKT or the OpenAir format), will the  library be maintained ?

babastienne (changes requested):
Hi. Thank you very much for your contribution !

Could you fix the codestyle of the code ? (you can follow the guidelines at https://geotrek.readthedocs.io/en/master/development.html#conventions to match our linting rules)

Also, because this is a specific feature that use an external library, we try to ensure that this will not introduce regressions in the future. The library PyOpenair does not seems to be tested at all and should be before we can validate the PR. Because if there is a regression we should be able to contribute to the project (and unit tests helps to understand use cases).

If a new standard appears in the future (for example in the WKT or the OpenAir format), will the library be maintained ?

lpofredc:
Replaced by #3349

PR #2741, opened 2021-08-31
Add download attachments to Suricate parser

Chatewgne:


geotrek/feedback/parsers.py
```diff
@@ -20,11 +22,17 @@
...
 
-class SuricateParser(AttachmentParserMixin, SuricateGestionRequestManager):
+class SuricateParser(SuricateGestionRequestManager):
 
     def __init__(self):
         super().__init__()
         self.bbox = Polygon.from_bbox(settings.SPATIAL_EXTENT)
+        # Replicated Attachment parser start
+        if settings.PAPERCLIP_ENABLE_LINK is False:
+            raise Exception('You need to enable PAPERCLIP_ENABLE_LINK to use this function')
```
submarcos, line 32: Il me semble que ce settings c'est pour stocker juste l'url de la piece jointe au lieu du fichier. du coup pourquoi l'exiger ?

submarcos: vu que plus loin tu telecharge la piece jointe

geotrek/feedback/parsers.py
```diff
@@ -32,7 +32,7 @@ def __init__(self):
         if settings.PAPERCLIP_ENABLE_LINK is False:
             raise Exception('You need to enable PAPERCLIP_ENABLE_LINK to use this function')
         try:
-            self.filetype = FileType.objects.get_or_create(type="Photographie", structure=None)
+            self.filetype, created = FileType.objects.get_or_create(type="Photographie", structure=None)
         except FileType.DoesNotExist:
```
gbip: First time reviewing, hi 👋
This is not 100% related to your change, but do you think that exception handling is really mandatory ?

geotrek/feedback/tests/test_suricate_sync.py
```diff
@@ -63,6 +70,11 @@ def build_response_patch(url, params=None, **kwargs):
             elif "GetAlerts" in url:
                 mock_response.content = mocked_json("suricate_alerts.json")
                 mock_response.status_code = 200
+            elif cause_JPG_error:
+                mock_response.status_code = 404
+            elif ".jpg" in url or ".png" in url or ".JPG" in url:
+                mock_response.content = mocked_image("theme-fauna.png")
+                mock_response.status_code = 200
```
gbip, line 77: Do you think that adding a http `content_type` header on top of the jpg data would improve your tests ?

geotrek/feedback/helpers.py
```diff
@@ -82,9 +82,23 @@ def post_to_suricate(self, endpoint, params):
...
+        if self.USE_AUTH:
+            response = requests.get(
+                url,
+                auth=self.AUTH,
+            )
+        else:
+            response = requests.get(
+                url,
+            )
         if response.status_code not in [200, 201]:
```
gbip, line 97: Should we handle [http 304 status code](https://httpwg.org/specs/rfc7232.html#status.304) ?
This status code is often found when there is some sort of caching in place.
If you know that your http client does transparent caching (it converts 304 response into 200 responses) then we don't need to handle it here.

PR #2742, opened 2021-09-02
APIv2 - Add routes for Service and ServiceType 

Chatewgne:


geotrek/api/v2/filters.py
```diff
@@ -452,6 +454,28 @@ def get_schema_fields(self, view):
...
 
 
+class GeotrekServiceFilter(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        qs = queryset
+        types = request.GET.get('types')
+        if types:
+            print(types)
+            types_id = types.split(',')
+            if ServiceType.objects.filter(id__in=types_id).exists():
```
submarcos, line 464: why test if service exists before filtering ?

PR #2747, opened 2021-09-06
Fix infrastructure difficulty levels translations and helptext display

Chatewgne:


geotrek/infrastructure/models.py
```diff
@@ -74,8 +73,8 @@ class InfrastructureUsageDifficultyLevel(StructureOrNoneRelated):
     label = models.CharField(verbose_name=_("Label"), unique=True, max_length=250)
 
     class Meta:
-        verbose_name = _("Infrastructure Usage Difficulty Levels")
-        verbose_name_plural = _("Infrastructure Usage Difficulty Levels")
+        verbose_name = _("Usage difficulty")
+        verbose_name_plural = _("Usage difficulty levels")
```
submarcos, line 77: i think you need to make a migration, it's required after changing verbose_name in models

submarcos: to prevent this prevent, add a ./manage.py makemigrations --check in circle ci file, in a separate job or just before launching tests

PR #2752, opened 2021-09-08
Fix missing condition in import sensitivity and tourism views

numahell:


submarcos:
ok for IF in import, but you should use IF in code too isn't it .. Is it working ?

numahell:
Curiously, `if` was already there when method using Sensitive model is called. I've been tested it in local, and sync_rando now works (was in error on master)

Chatewgne (approved):

PR #2803, opened 2021-10-22
Add option reload path altimetry after load_dem

LeGrandTim:


geotrek/altimetry/management/commands/loaddem.py
```diff
@@ -19,11 +19,14 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('dem_path')
         parser.add_argument('--replace', action='store_true', default=False, help='Replace existing DEM if any.')
+        parser.add_argument('--update-paths', action='store_true', default=False,
```
submarcos, line 22: yes, you update path to update 3d topologies, but the goal is not to update path. I suggest --update-3d or --update-topologies-3d or other thing

submarcos: --update-altimery ?

camillemonchicourt (approved):

PR #2814, opened 2021-11-08
Fix generation altimetry profile

LePetitTim:
Issue with the generation of altimetry profiles :
Capture_d_écran_de_2021-11-02_09-39-42
The generation used a matrix to get every altimetry for each points on the area every x meters.
The problem was related to the way we crossed the array before generate the points. We can cross the array, firstly increasing x then y or the opposite. 

After that we create a matrix of all elevations, but the matrix is not the same depending on the first array :

2 differents array order:

1.  [[x0, y0], [x0, y1], [x0, y2], [x1, y0], [x1, y1], [x1, y2]] 
2.  [[x0, y0], [x1, y0], [x0, y1], [x1, y1], [x0, y2], [x1, y2]]

- x0,y0 = z0
- x0, y1 = z1
- x0, y2 = z2
- x1, y0 = z3
- x1, y1 = z4
- x1, y2 = z5

Generate an array of elevations depending on the length of all Xs (2)
 1. [[z0, z1], [z2, z3], [z4, z5]]
 2. [[z0, z3], [z1, z4], [z2, z5]]
 
 We had to order the query to get all the time the same result.

### DIFF
--- a/geotrek/altimetry/tests/test_elevation.py
+++ b/geotrek/altimetry/tests/test_elevation.py
@@ -230,6 +230,52 @@ class ElevationOtherGeomAreaTest(AreaTestCase):
         self.assertEqual(area['size']['y'], 1300.0)
 
 
+class ElevationRightOrderAreaTest(AreaTestCase):
+    def _fill_raster(self):
+        with connection.cursor() as cur:
+            cur.execute('INSERT INTO altimetry_dem (rast) VALUES (ST_MakeEmptyRaster(250, 250, 0, 250, 25, -25, 0, 0, %s))', [settings.SRID])
+            cur.execute('UPDATE altimetry_dem SET rast = ST_AddBand(rast, \'16BSI\')')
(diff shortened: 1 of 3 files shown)


Chatewgne (approved):

Chatewgne (approved):

geotrek/altimetry/tests/test_elevation.py
```diff
@@ -230,6 +230,51 @@ def test_area_has_nice_ratio_if_square_enough(self):
         self.assertEqual(area['size']['y'], 1300.0)
 
 
+class ElevationOtherAreaTest(AreaTestCase):
```
submarcos, line 233: other quoi ? qu'est-ce que ce test teste de différent ?

LePetitTim: J'ai changé pour ElevationRightOrderAreaTest

PR #2872, opened 2021-12-24
Fix bootstrap messages (XMas PR 🎅)

submarcos:
Fix conf to use right bootstrap tags in alert and messages in geotrek / mapentity.

Capture d’écran du 2021-12-24 16-10-40

Transparent background will be replaced by yellow for warning and light red for errors


camillemonchicourt:
Thanks 🎅

numahell (approved):

PR #2911, opened 2022-01-19
Workflow - Send new reports email to Workflow Manager instead of regular managers

Chatewgne:


geotrek/feedback/tests/test_suricate_sync.py
```diff
@@ -119,6 +119,12 @@ def build_extra_failed_request_patch(self, mocked: MagicMock):
         mock_response.content = {}
         mocked.return_value = mock_response
 
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = UserFactory(username="Maxou", password="drowssap")
```
submarcos, line 124:
```suggestion
        cls.user = UserFactory()
```
I think password and username doesn't mater here

submarcos (approved):

numahell (approved):

PR #2929, opened 2022-01-25
Serve 3D in API v2

Chatewgne:


submarcos:
geotrek.altimetry is always installed, and I think ``dem`` should be an extra detail action on TrekViewset

```python
class TrekViewSet(xxx):
    ....
    @action(detail=True)
    def dem(....):
         return Response(self.get_object().dem))
```

PR #2946, opened 2022-02-03
Change equipment gear in trekking, move order, fix api v2 translate e…

LePetitTim:
…quipment


geotrek/api/v2/serializers.py
```diff
@@ -1056,6 +1060,7 @@ class CourseSerializer(PDFSerializerMixin, DynamicFieldsMixin, serializers.Model
         parents = serializers.ReadOnlyField(source='parents_id')
         accessibility = serializers.SerializerMethodField()
         attachments = AttachmentSerializer(many=True, source='sorted_attachments')
+        equipment = serializers.SerializerMethodField()
```
submarcos, line 1063: I don't understand here why you delete equipment in fields but add attribute

geotrek/api/v2/serializers.py
```diff
@@ -1069,6 +1074,9 @@ def get_accessibility(self, obj):
         def get_cities(self, obj):
             return [city.code for city in obj.published_cities]
 
+        def get_equipment(self, obj):
+            return get_translation_or_dict('equipment', self, obj)
```
submarcos, line 1078: idem

geotrek/api/v2/serializers.py
LePetitTim: This is not for the same model it is for outdoor course (we should translate it)

geotrek/api/v2/serializers.py
LePetitTim: This is not for the same model it is for outdoor course (we should translate it)

LePetitTim: - Fix translation equipment api v2 courses

PR #2973, opened 2022-03-04
Remove SVG sizes in pictograms

camillemonchicourt:
As width and height in SVG can generate problems in Geotrek-rando-v3


submarcos:
Before merging we need to check display behaviour in Geotrek-rando v2

camillemonchicourt:
I checked about it.
It should be OK as I did the same with my pictograms that are used in our Geotrek-rando-v3 (https://rando-ecrins-parcnational.fr), but also some Geotrek-rando-v2 : 
- http://www.grand-tour-ecrins.fr/?categories=T2&categories=T3&categories=T4&categories=T8
- http://www.rando-serreponcon.com/?categories=T4
- https://rando.paysdesecrins.com/?categories=T4
- https://rando-brianconnais.fr/?categories=T2&categories=T4&categories=T8

And in any cases, these changes to pictograms will only apply to new Geotrek-admin installations. Those will not use v2 of Geotrek-rando.

PR #3003, opened 2022-03-15
Add filter exclude labels

LePetitTim:


camillemonchicourt (approved):

submarcos (commented):
I think you should add translations

LePetitTim:
> I think you should add translations

indeed

Chatewgne (approved):

submarcos (approved):

PR #3023, opened 2022-03-21
Refer to Reports by a label instead of email addresses

Chatewgne:


geotrek/feedback/models.py
```diff
@@ -98,19 +98,20 @@ class Meta:
...
 
     @property
-    def email_display(self):
-        return '<a data-pk="%s" href="%s" title="%s" >%s</a>' % (
-            self.pk,
-            self.get_detail_url(),
-            self,
-            self,
-        )
+    def id_display(self):
```
submarcos, line 109: for mapentity compatibility you should not customize id display but add other named property

geotrek/feedback/tests/test_commands.py
```diff
@@ -27,7 +27,7 @@ def test_erase_old_emails(self):
         call_command('erase_emails', stdout=output)
         old_report = Report.objects.get(id=self.old_report.id)
         self.assertEqual(old_report.email, "")
-        self.assertEqual(old_report.__str__(), "Anonymous report")
+        self.assertEqual(old_report.__str__(), f"{_('Report')} n°{old_report.pk}")
```
submarcos, line 30: I think tests are run in english only (default language)

geotrek/feedback/views.py
```diff
@@ -38,7 +38,7 @@ class ReportList(CustomColumnsMixin, mapentity_views.MapEntityList):
     )
     model = feedback_models.Report
     filterform = ReportFilterSet
-    mandatory_columns = ['id', 'email', 'activity']
+    mandatory_columns = ['id', 'id', 'activity']
```
submarcos, line 41: I f you use id two times, you duplicate column. you should not use ID as label, use a  label property with custom _display

PR #3026, opened 2022-03-22
Precise formats handled by `loadXX` commands

IdrissaD:
If I understand correctly, all GDAL API-handled formats can be imported via `loadpoi`, `loadsignage` (etc.) commands or Geotrek-admin UI. It's useful to know it, so correct the PR with the correct formats if i'm wrong.

### DIFF
--- a/docs/import.rst
+++ b/docs/import.rst
@@ -116,7 +116,7 @@ Open the top right menu and clic on ``imports``.
 Import from a file
 ------------------
 
-You can also use some of Geotrek commands to import data from a file.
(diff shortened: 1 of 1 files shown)


submarcos:
> If I understand correctly, all GDAL API-handled formats can be imported via `loadpoi`, `loadsignage` (etc.) commands or Geotrek-admin UI. It's useful to know it, so correct the PR with the correct formats if i'm wrong.

You're true, but only vector formats depending on gdal version installed. I suggest to precise https://gdal.org/index.html instead of django documentation, and ``gdal-config --formats`` command to get supported formats on installed version (not the same on bionic or focal for example)

submarcos:
bionic is gdal 2.2.3, focal (and docker images based on it) is 3.0.4

IdrissaD:
Thanks, it's corrected

camillemonchicourt (approved):

PR #3033, opened 2022-03-25
Fix template pdfs map/image, accessibility, elevation

LePetitTim:


LePetitTim:
Before : 

After :

LePetitTim:
Before :


After :


Was working only when departure or arrival was long enough

camillemonchicourt (approved):

dtrucs (approved):

submarcos (approved):

PR #3052, opened 2022-04-04
Add doc for control characters count in TinyMCE fields

numahell:
This PR allows to control characters count in TinyMCE fields, using max characters count setting `MAX_CHARACTERS` (default = 0)

Can ben merge when this PR is merged and a new release for Mapentity https://github.com/makinacorpus/django-mapentity/pull/237 for details.


numahell:
It will work on all text fields : I backported code in django mapentity, we think it can be used in other projects.

camillemonchicourt:
Yes it is an interesting feature.
But I think that people will want to limit the number of characters for texts from trekking and maybe outdoor (to be displayed in Geotrek-rando), but certainly not for maintenance, signage and maintenance text fields, where there is no reason to limit the number of characters.

If this setting is global, I am not sure it will be usable.

Chatewgne (approved):

PR #3068, opened 2022-04-20
Change filter sensitive add distance SENSITIVE_AREA_INTERSECTION_MARGIN

LePetitTim:


geotrek/api/v2/filters.py
```diff
@@ -411,6 +413,10 @@ def filter_queryset(self, request, queryset, view):
         types = request.GET.get('types')
         if types:
             qs = qs.filter(type__in=types.split(','))
+        trek = request.GET.get('trek', None)
+        if trek is not None:
+            t = Trek.objects.get(pk=trek)
```
Chatewgne, line 418: Use `filter` instead of `get` to avoid exceptions when bad trek ID is provided

camillemonchicourt:
Not sure to understand changelog. This is a filter added in API v2 ? On which route ?

LePetitTim:
> Not sure to understand changelog. This is a filter added in API v2 ? On which route ?

Is it better with : 
- Use distance from setting SENSITIVE_AREA_INTERSECTION_MARGIN in sensitive area filter `trek`

geotrek/api/v2/filters.py
LePetitTim: Done

camillemonchicourt:
Yes, but should mention that it is in API v2, as we could understand that it is in Geotrek-admin.

PS : Just saw that Sensitive area route as two similar filters that can bring confusion : "trek" and "near_trek"...

Chatewgne (approved):

PR #3087, opened 2022-05-04
Docs import paths

numahell:


docs/install/import.rst
```diff
@@ -2,6 +2,53 @@
 Import data
 ===========
 
+Import paths
+============
+
+.. danger::
+    With dynamic segmentation, importing paths is only possible if no paths are created yet.
```
Chatewgne:
```suggestion
    With dynamic segmentation, importing paths is only safe if no paths are created yet.
```

docs/install/import.rst
```diff
@@ -2,6 +2,53 @@
...
+
+    If you import paths where there are existing paths or treks, their linked topology will be modified automatically.
```
Chatewgne, line 11:
```suggestion
    If you import paths where there are existing paths or treks, their linked topology might be impacted.
```

numahell: J'ai un peu modifié la phrase

docs/install/import.rst
numahell: J'ai un peu modifié la phrase

docs/install/import.rst
```diff
@@ -2,6 +2,54 @@
 Import data
 ===========
 
+Import paths
+============
+
+.. danger::
+    With dynamic segmentation, importing paths is very risky if paths are already present in Geotrek,
+    it is only safe with no path created.
```
Chatewgne, line 10:
```suggestion
    it is only safe with no pre-existing paths.
```

docs/install/import.rst
```diff
@@ -2,6 +2,54 @@
...
+
+    Indeed, if you import paths where there are existing paths, treks, POIs or trails linked topology might be impacted.
+
+Before import paths layer, it is important to prepare them. Paths must be:
```
Chatewgne, line 14:
```suggestion
Before importing paths layer, it is important to prepare them. Paths must be:
```

Chatewgne (approved):

docs/install/import.rst
```diff
@@ -2,6 +2,54 @@
 Import data
 ===========
 
+Import paths
+============
+
+.. danger::
+    With dynamic segmentation, importing paths is very risky if paths are already present in Geotrek,
```
camillemonchicourt, line 9: I would add "in the same area". Because importing some paths in an area without pre-existing ones, is not a problem. 
I also import paths all the time in areas that already have paths, I just take care in QGIS to msnap new paths with pre-existing ones. 
I use this method : https://makina-corpus.com/sig-webmapping/importer-une-couche-de-troncons-dans-geotrek

So I may wouldn't be so scary. 
But maybe with the UI import tool, it is more complicated and more risky. 

The good way to add paths in a Geotrek with existing one, is to prepare them in QGIS with displaying the path layer from Geotrek to snap new paths to existing ones, and to never import a path overriding an existing one.

numahell: Ah yes good point, and I've forgottent this article from Mathieu :)

numahell: Hello, I've made some improvements with your feedback

camillemonchicourt (approved):

PR #3112, opened 2022-05-23
Add an acitivity in feedback basic fixture

Chatewgne:


numahell (approved):

geotrek/feedback/fixtures/basic.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "model": "feedback.reportactivity",
+        "pk": 18,
```
submarcos, line 4: I'm not sure but with postgresql serial auto increment, I think use 18 as first value is not a good idea. Use 1 instead ?

submarcos (approved):

PR #3187, opened 2022-07-25
Circleci update resource class

submarcos:


.circleci/config.yml
```diff
@@ -9,149 +9,273 @@ jobs:
...
+          command: BASE_IMAGE_TAG=focal-3.8 make build
+      - run:
+          name: Tag and save docker image
+          command: |
+            docker tag geotrek:latest geotrekce/circleci-admin:$CIRCLE_WORKFLOW_ID
+            docker save -o ./20.04.$CIRCLE_WORKFLOW_ID.tar geotrekce/circleci-admin:$CIRCLE_WORKFLOW_ID
+      - persist_to_workspace:
+          root: .
+          paths:
+            - 20.04.*.tar
```
LePetitTim, line 60: Pourquoi tu reutilises pas $CIRCLE_WORKFLOW_ID ?

.circleci/config.yml
```diff
@@ -9,149 +9,273 @@ jobs:
...
+    machine:
+      image: ubuntu-2004:202201-02
     environment:
       LANG: C.UTF-8
       ENV: tests
       SECRET_KEY: foo
       CUSTOM_SETTINGS_FILE: ""
       CONVERSION_HOST: localhost
       CAPTURE_HOST: localhost
-    working_directory: /opt/geotrek-admin
```
LePetitTim, line 119: Why did it exist before ?

.circleci/config.yml
submarcos: no matter WORKFLOW ID because persist_to_workspace is unique by workflow

submarcos: and it's not possible to use ENV variable here

.circleci/config.yml
submarcos: before, previous built docker image was used directly by CIrcle. Now, we load image and launch it. 
In bot cases, /opt/geotrek-admin is always default working_directory

.circleci/config.yml
submarcos: fixed by using $CIRCLE_SHA1 (allow to relaunch failed job) and not specify ID in saved image

PR #3233, opened 2022-09-13
Fix parameter name MAIL_MANAGER in documentation

numahell:


docs/changelog.rst
```diff
@@ -14,6 +14,10 @@ CHANGELOG
 
 - Fix Site creation form is initialized with parent Site
 
+**Documentation**
+
+- Fix parameter name ``MAIN_MANAGERS`` in documentation
```
camillemonchicourt, line 19: Should be MAIL_MANAGERS

docs/install/advanced-configuration.rst
```diff
@@ -48,7 +48,7 @@ Geotrek-admin will send emails:
 * to managers when a feedback report is created
 
 Email configuration takes place in ``/opt/geotrek-admin/var/conf/custom.py``, where you control
-recipients emails (``ADMINS``, ``MAIL_MANAGERS``) and email server configuration.
+recipients emails (``ADMINS``, ``MANAGERS``) and email server configuration.
```
camillemonchicourt, line 51: We assume that administrators will understand how to fill these parameters or will check examples in https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/settings/custom.py.dist ?

docs/changelog.rst
submarcos, line 19:
```suggestion
- Fix parameter name ``MAIL_MANAGERS`` in documentation
```

PR #3234, opened 2022-09-13
Add missing changelog

numahell:


camillemonchicourt:
Yes it was included in 2.86.0.
So it should also be removed from 2.85.0 version in which it was not included.

Chatewgne:
Intégré à la main dans le changelog de la release 2.87.0

PR #3236, opened 2022-09-16
Fix synchro rando after fresh install on Debian

marcantoinedupre:
root owned installed media files from fixtures. Sync rando ran from the web UI (with geotrek system user) could not process those files resulting in a PermissionError.


submarcos (commented):
add changelog and OK for me

PR #3240, opened 2022-09-21
Fixes for GeotrekAggregator

Chatewgne:
- Fix provider checking for TouristicContentParser
- Add documentation on using `provider` in Parsers
- Improve translation management in Parsers


camillemonchicourt (approved):

geotrek/common/parsers.py
```diff
@@ -243,8 +256,8 @@ def parse_field(self, row, dst, src, updated, non_field):
         if modified:
             updated.append(dst)
             if dst in self.translated_fields:
-                lang = translation.get_language()
-                updated.append('{field}_{lang}'.format(field=dst, lang=lang))
+                for lang in settings.MODELTRANSLATION_LANGUAGES:
+                    updated.append('{field}_{lang}'.format(field=dst, lang=lang))
```
Chatewgne, line 261: Only `name` and `name_fr` would be updated, now all translations can be updated

geotrek/common/parsers.py
```diff
@@ -202,24 +206,33 @@ def set_value(self, dst, src, val):
...
         else:
             val = self.apply_filter(dst, src, val)
         if hasattr(self.obj, dst):
             if dst in self.m2m_fields or dst in self.m2m_constant_fields:
-                old = set(getattr(self.obj, dst).all())
                 val = set(val)
                 if dst in self.m2m_aggregate_fields:
                     val = val | old
-            else:
-                old = getattr(self.obj, dst)
```
Chatewgne, line 216: We need to retrieve `old` value BEFORE running `apply_filter` because this method could happen to change the value on the object (like in Aggregator)

geotrek/common/parsers.py
```diff
@@ -113,6 +115,8 @@ def __init__(self, progress_cb=None, user=None, encoding='utf8'):
                 f.name: force_str(f.verbose_name)
                 for f in self.model._meta.many_to_many
             }
+        default_language = self.default_language if self.default_language else settings.MODELTRANSLATION_DEFAULT_LANGUAGE
+        translation.activate(default_language)
```
Chatewgne, line 119: Use `translation.activate` to make sure `name` field is properly linked to the right language `name_fr` field (otherwise we could happen to impact the wrong translation field when using `object.name = "Nom fr"` or `setattr(object, 'name', "Nom fr")`

submarcos (approved):

PR #3247, opened 2022-09-27
add bookable field

TheoLechemia:
UI : make related field dynamic


geotrek/api/v2/filters.py
```diff
@@ -535,6 +535,14 @@ def get_schema_fields(self, view):
                     title=_("Dates after"),
                     description=_("Filter events happening after or during date, format YYYY-MM-DD")
                 )
+            ), Field(
+                name='bookable',
```
Chatewgne, line 539: This causes an error because fields added with `filterset_fields` are automatically added to Swagger

Please remove it from this method and find a way to set description on the automatically generated one : 

image

Chatewgne (approved):

geotrek/api/v2/views/tourism.py
```diff
@@ -84,8 +84,9 @@ class TouristicEventViewSet(api_viewsets.GeotrekGeometricViewset):
     filter_backends = api_viewsets.GeotrekGeometricViewset.filter_backends + (
         api_filters.GeotrekTouristicEventFilter,
         api_filters.NearbyContentFilter,
-        api_filters.UpdateOrCreateDateFilter
+        api_filters.UpdateOrCreateDateFilter,
     )
+    filterset_fields = ["bookable"]  # TODO find a way to overide automatic Swagger doc to add a description for this filter
```
submarcos, line 89: Many ways to add doc : 
  - default is model field help_text
  - specific for API : use filterset_class instead of filterset_fields, inheritad from djang_filters.FilterSet. and customize bookable field with specific help_text

submarcos (commented):
all translation files should be updated, not only french

PR #3251, opened 2022-09-28
Test begin date is before end date

amandine-sahl:
Test begin date is lower than end date


geotrek/tourism/locale/fr/LC_MESSAGES/django.po
```diff
@@ -20,6 +20,10 @@ msgstr ""
 "X-Generator: Weblate 4.0.1\n"
 "Project-Id-Verésion: PACKAGE VERSION\n"
 
+
+msgid "Error date start > end"
+msgstr "La date de début est suppérieur à la date de fin"
```
camillemonchicourt, line 25: supérieure

PR #3257, opened 2022-09-28
Formset layout

bouttier:


Chatewgne:
Please add tests for missing lines
Add cypress test for formsets ?

Chatewgne:
What about `Fieldset(_('Lines'))` in `signage/forms.py` ?

bouttier:
There is still a display behaviour on the trek update page when using Formset for trek relationships

PR #3261, opened 2022-09-29
Feat/eventparticipantnumber

amandine-sahl:
- [x] renommer champ participant_number en capacity
- [x] changer type en integer
- [x] essayer de transformer les données de participant_number si integer, sinon les mettre dans le champ practical_info
- [x] test parseur apidae


geotrek/tourism/parsers.py
```diff
@@ -374,6 +374,12 @@ def filter_contact(self, src, val):
         ] if line]
         return '<br>'.join(lines)
 
+    def filter_capacity(self, src, val):
+        if val.isnumeric():
+            return int(val)
+        else:
+            return None
```
Chatewgne, line 381: Use `self.add_warning()` to output a warning stating that capacity received a bad value

PR #3269, opened 2022-09-30
Feat/eventlegacy

amandine-sahl:
Add rename fields in event api v2 :
- [x] create LegacyFieldMixin
- [x] meeting_time -> start_time
- [ ] participant_number -> capacity


camillemonchicourt:
Finally we choose to display old field name along with new name, but to indicate that the old field name is depecrated.

PR #3275, opened 2022-10-01
Order touristic event form and detail fields

amandine-sahl:
Change field order :

- [x] detail
- [x] form
- [x] export
- [x] documents (pdf, ...)


camillemonchicourt (approved):

Chatewgne:
Rebase and add `place` field as well

PR #3280, opened 2022-10-04
Fixes

TheoLechemia:
Translation and missing place input


geotrek/tourism/locale/en/LC_MESSAGES/django.po
```diff
@@ -329,8 +329,8 @@ msgstr ""
 msgid "Boolean indicating if Event is cancelled"
 msgstr ""
 
-msgid "Select a place to auto-locate event on map"
-msgstr ""
+msgid "Select a place on the list or locate the event directly on the map"
```
camillemonchicourt, line 332: It should be "in the list", not "on the list".

PR #3281, opened 2022-10-04
add preparation_duration and intervention_duration

TheoLechemia:
Add 2 new integer fields : 
- preparation_duration
- intervention_duration
These fields are "private" so I don't add them in API


geotrek/tourism/models.py
```diff
@@ -477,6 +477,8 @@ class TouristicEvent(ZoningPropertiesMixin, AddPropertyMixin, PublishableMixin,
     bookable = models.BooleanField(verbose_name=_("Bookable"), default=False)
     cancelled = models.BooleanField(default=False, verbose_name=_("Cancelled"), help_text=_("Boolean indicating if Event is cancelled"))
     cancellation_reason = models.ForeignKey(CancellationReason, verbose_name=_("Cancellation reason"), related_name="touristic_events", null=True, blank=True, on_delete=models.PROTECT)
+    preparation_duration = models.IntegerField(verbose_name=_("Preparation duration"), blank=True, null=True)
+    intervention_duration = models.IntegerField(verbose_name=_("Intervention duration"), blank=True, null=True)
```
Chatewgne, line 481: If this is meant to be in Hours I think we should copy this field : https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/trekking/models.py#L135
with FloatField, validator and help_text

docs/install/advanced-configuration.rst
```diff
@@ -2776,7 +2780,9 @@ An exhaustive list of form fields hideable in each module.
             "bookable",
             'cancelled',
             'cancellation_reason'
-            'place'
+            'place',
+            'preparation_duration',
+            'intervention_duration'
```
Chatewgne, line 2785: Please update `docs/changelog.rst` with these new fields

Chatewgne (changes requested):

geotrek/tourism/models.py
```diff
@@ -536,31 +759,75 @@ def meta_description(self):
...
+)
+TouristicEvent.add_property(
+    "touristic_events",
+    lambda self: intersecting(TouristicEvent, self),
+    _("Touristic events"),
+)
+TouristicEvent.add_property(
+    "published_touristic_events",
+    lambda self: intersecting(TouristicEvent, self).filter(published=True),
+    _("Published touristic events"),
```
Chatewgne, line 832: Please undo whole file formatting

camillemonchicourt:
@TheoLechemia, il faut aussi faire pointer cette PR vers la branche "master".

geotrek/tourism/models.py
TheoLechemia: DOne

TheoLechemia:
Do I have to modify the PDF ?

Chatewgne:
> Do I have to modify the PDF ?

I don't think so, @camillemonchicourt ?

camillemonchicourt:
Not it shouldn't be in the public PDF, as these are private fields that we don't put in API v2, nor Geotrek-rando, nor the PDF.

marcantoinedupre (approved):
It looks good! :+1:

PR #3287, opened 2022-10-07
Add direction on signage line

marcantoinedupre:
This is the PR to add a mode where the directions are on the lines rather than the blades in the `signage` app. This is driven by the `DIRECTION_ON_LINES_ENABLED` setting.

I split the PR into small commits for an easy review. I could squash some of them before merging if convenient.


camillemonchicourt:
Related to https://github.com/GeotrekCE/Geotrek-admin/issues/3178

geotrek/common/viewsets.py
```diff
@@ -6,9 +6,10 @@ class GeotrekMapentityViewSet(MapEntityViewSet):
     """ Custom MapentityViewSet for geotrek. """
 
     permission_classes = [permissions.DjangoModelPermissionsOrAnonReadOnly]
+    mapentity_list_class = None
```
submarcos, line 9:
```suggestion
    mapentity_list_class = []
```
I suggest to keep same type as empty list by default

numahell (approved):

Chatewgne (approved):

Issue #1815, opened 2018-12-14
Add languages in parsers


LePetitTim:

camillemonchicourt:
Yes, would be useful for APIDAE parser as an example as APIDAE data are available in several language but we import only one.

LePetitTim:
Exactly.

camillemonchicourt:
New PR in progress : https://github.com/GeotrekCE/Geotrek-admin/pull/2308

camillemonchicourt:
First part included in version 2.89.0 : Partially handle translated fields: when setting ``fill_empty_translated_fields`` to ``True``, all empty translation fields for all languages will be set with the parsed value

PR #3301, opened 2022-10-17
Parsers multilang

numahell:
Add method to handle translated fields.
When parser setting `fill_empty_translated_fields` is True (default is False), parser will update empty translated fields with value provided by the source flow or file (exemple, if `name_it` is empty and not default language, it will be filled with value provide by source flow whatever the lang).
This PR only concerns source flow with a unique field, next step is to handle source with a field by language.


geotrek/common/parsers.py
```diff
@@ -77,6 +77,7 @@ class Parser:
     update_only = False
     delete = False
     duplicate_eid_allowed = False
+    fill_empty_translated_fields = False
```
Chatewgne, line 80: Keep documenting new attributes above

geotrek/common/parsers.py
```diff
@@ -254,6 +275,8 @@ def parse_field(self, row, dst, src, updated, non_field):
             val = self.get_val(row, dst, src)
         if non_field:
             modified = self.parse_non_field(dst, src, val)
+        elif self.fill_empty_translated_fields and dst in self.translated_fields:
+            modified = self.parse_translated_field(dst, src, val)
```
marcantoinedupre, line 279: By shunting out `self.parse_real_field` in the case of a translated field the `self.apply_filter` will not be call if a filter exists for that field.
I don't know if translated fields often have filters but not calling it breaks the parser filter hook interface.
I don't see a better way than just call it in `parse_translated_field`, maybe by wrapping the check for a filter method and the call in a new method. It would be better to add a test case with a filtered translated fields too.

marcantoinedupre (commented):
Nice work :)

geotrek/common/parsers.py
numahell: Yes, I tried to list all basic settings for parsers, it was not present in documentation.

geotrek/common/parsers.py
numahell: Exact, thanks for having seen this problem.

camillemonchicourt:
Related to https://github.com/GeotrekCE/Geotrek-admin/issues/1815

geotrek/common/parsers.py
numahell: Normally fixed

docs/changelog.rst
```diff
@@ -13,6 +13,8 @@ CHANGELOG
 **New features**
 
 - Add new setting `DIRECTION_ON_LINES_ENABLED` to have the `direction` field on lines instead of blades
+- Partially handle translated fields, when setting ``fill_empty_translated_fields`` is True,
+  all empty translated field will be set with value
```
marcantoinedupre, line 17: Proposed rephrasing : "Partially handle translated fields: when setting ``fill_empty_translated_fields`` to True,
  all empty translation fields for all languages will be set with the parsed value"
The `django-modeltranslation` lib names the `*_fr`, `*_en` fields _translation fields_. I think it would be nice to keep that to make it clear when we talk about the translated field (for instance `label`) and the translation fields (`label_fr`, `label_en`, etc).

marcantoinedupre (approved):

PR #3315, opened 2022-11-02
Add Trek parser from APIDAE

marcantoinedupre:
This PR adds a new Parser subclass to import Treks from the APIDAE touristic data system.

The specificities of that parser are:

- It can import APIDAE translated fields into Geotrek translated fields (more details below).
- Default filtering/mapping on relevant APIDAE reference elements (APIDAE themes which are relevant to categorize treks, labels, ref elements for a difficulty scale, etc).
- Linked entities (Themes, Labels) are also imported through secondary parsers using the configured filtering/mapping of APIDAE reference elements.

schema_ApidaeTrekParser

## Simple import of 1 APIDAE Itinéraire translated field into 1 Geotrek Trek translated field
For instance : `"nom.libelleFr", "nom.libelleEn", etc -> "name.fr", "name.en", etc`.
New field option for that parser `expand_translations`.

## Import multiple APIDAE Itinéraire translated fields into 1 Geotrek Trek translated field

For instance the Trek's description (step-by-step guidebook, closed temporarily, opening info, is the trek marked?, ...). It is done through a specific implementation of the filter hook (for instance `filter_description`) which prepares an instance of TranslatedField with all concatenated translations. The `apply_filter` is overridden to dispatch TranslatedField values to the translation fields of the Trek. `apply_filter` is called at the end of relevant filter hooks.

## Import translations of linked entities

For instance translations of a Theme -> done with a secondary parser requesting the APIDAE endpoint `/referentiel/elements-reference` with the ID list of all relevant entities. The fields mapping is generated in the Reference Element parser's `__init__`.

Implemented secondary parsers:

```
ApidaeReferenceElementParser > ApidaeTrekThemeParser
                               > ApidaeTrekLabelParser
                               > ApidaeTrekDifficultyParser
                               > ApidaeTrekNetworkParser
                               > ApidaeTrekAccessibilityParser
```
                  
## Limitations

- translated fields imported with TranslatedField won't mark the Trek as updated when only the secondary languages have changed but not the default one. We would need to modify the core parser mechanisms to fix this.

## FIXMES?

- In order to have a custom behavior on a specific Geotrek instance I added a hook method `ApidaeTrekParser._transform_guidebook_to_html` which does nothing but redirects to `ApidaeTrekParser._transform_description_to_html`. I do not like it because from the main codebase point of view you cannot see what is its purpose. How could we improve that?
- I had to use a TemporaryNamedFile to use the GDAL DataSource to import the GPX data. I don't think that can be avoided, can it?
- I am not sure of how to deal with parsing errors. I could probably be improved.


Chatewgne (changes requested):

geotrek/common/models.py
```diff
@@ -178,7 +178,7 @@ def pictogram_off(self):
 
 
 class RecordSource(OptionalPictogramMixin):
-    name = models.CharField(verbose_name=_("Name"), max_length=50)
+    name = models.CharField(verbose_name=_("Name"), max_length=80)
```
LePetitTim, line 181: Why 80 ?

geotrek/trekking/parsers.py
```diff
@@ -208,3 +216,398 @@ class GeotrekPOIParser(GeotrekParser):
...
+        source.website = manager['siteWeb']
+        source.save()
+        return sources
+
+    def filter_difficulty(self, src, val):
+        types_clientele = val
+        difficulty_level = None
+        for tc in types_clientele:
+            if tc['id'] in TYPES_CLIENTELE_IDS_AS_DIFFICULTY_LEVELS:
+                difficulty_level = tc
```
LePetitTim, line 448: you could return here ?

```
self.apply_filter(
                dst='difficulty',
                src=src,
                val=difficulty_level[f'libelle{settings.MODELTRANSLATION_DEFAULT_LANGUAGE.capitalize()}']
            )
            ```

File: geotrek/common/models.py (hunk 1 above)
Comment (marcantoinedupre) [reply geotrek/common/models.py:181]: 50 is the current max length and it is slightly too short for some names. IMO increasing the length to 80 characters looks acceptable to accommodate APIDAE users.

File: geotrek/tourism/parsers.py [hunk 3]
```diff
@@ -98,34 +89,6 @@ class ApidaeParser(AttachmentParserMixin, Parser):
...
-            for row in self.items:
-                yield row
-            self.skip += self.size
-            if self.skip >= self.nb:
-                return
-
-    def normalize_field_name(self, name):
-        return name
-
     def filter_eid(self, src, val):
```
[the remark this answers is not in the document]
marcantoinedupre: I tried to move only the behavior needed by the `ApidaeTrekParser` which is the way to build the URL and to request the API. I could move the whole class but `ApidaeTrekParser` would inherit attributes or filter hooks it does not need.
What troubles you in the fact of splitting this class?

geotrek/trekking/parsers.py
```diff
@@ -208,3 +216,398 @@ class GeotrekPOIParser(GeotrekParser):
...
+
+
+class ApidaeTrekDifficultyParser(ApidaeReferenceElementParser):
+    model = DifficultyLevel
+    element_reference_ids = TYPES_CLIENTELE_IDS_AS_DIFFICULTY_LEVELS
+    name_field = 'difficulty'
+
+
+# TODO
+# class ApidaeTrekAccessibilityParser(ApidaeReferenceElementParser):
```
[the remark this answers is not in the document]
marcantoinedupre: Thanks for pointing it out. So far there are no accessbility-related attachments to import though.

geotrek/trekking/parsers.py
marcantoinedupre: I do not understand. Did you mean it misses an explicit return when `difficulty_level` is None?

marcantoinedupre: OK, the proposition is to return from inside the loop as soon as I find the matching difficulty level. Though shorter I don't like to return from inside clauses. ;)

geotrek/trekking/parsers.py
```diff
@@ -208,3 +216,398 @@ class GeotrekPOIParser(GeotrekParser):
...
+    }
+    non_fields = {}
+
+    def __init__(self, *args, **kwargs):
+        self._translated_fields = [field for field in get_translated_fields(self.model)]
+        self._expand_fields_mapping_with_translation_fields()
+        self._related_treks_mapping = defaultdict(list)
+        super().__init__(*args, **kwargs)
+
+    def _expand_fields_mapping_with_translation_fields(self):
```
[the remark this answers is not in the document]
marcantoinedupre: Maybe later when another use case appears.

geotrek/tourism/parsers.py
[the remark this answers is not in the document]
marcantoinedupre: OK, good point for the forgotten `filter_eid`. I will not place it in the base class as it is related to field mapping which is the concern of subclasses here. But I have added it to `geotrek.trekking.parsers.ApidaeTrekParser` just in case.

geotrek/trekking/parsers.py
```diff
@@ -208,3 +216,377 @@ class GeotrekPOIParser(GeotrekParser):
...
+
+    def end(self):
+        self._finalize_related_treks_association()
+
+    def filter_geom(self, src, val):
+        plan = self._find_gpx_plan_in_multimedia_items(val)
+        gpx = self._fetch_gpx_from_url(plan)
+
+        # FIXME: is there another way than the temporary file? It seems not. `DataSource` really expects a filename.
+        with NamedTemporaryFile(mode='w+b', dir='/opt/geotrek-admin/var/tmp') as ntf:
```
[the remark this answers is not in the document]
marcantoinedupre: Good point, done.

geotrek/trekking/tests/test_parsers.py
```diff
@@ -563,3 +573,543 @@ def test_create(self, mocked_head, mocked_get):
...
+
+        self.assertTrue(trek.practice is not None)
+        self.assertEqual(trek.practice.name, 'Pédestre')
+
+        self.assertEqual(trek.networks.count(), 2)
+        networks = trek.networks.all()
+        self.assertIn('Hiking itinerary', [n.network for n in networks])
+        self.assertIn('Pedestrian sports', [n.network for n in networks])
+
+        from geotrek.common.models import Attachment
```
Chatewgne, line 682: Is there a reason not to have this at the top ?

geotrek/trekking/tests/test_parsers.py
```diff
@@ -110,6 +117,9 @@ def test_multilinestring_with_hole(self):
     b'356612.3778 6689511.5083, 356447.7689 6689502.3634)'
 )
 
+# return # Produce a small red dot
+IMG_FILE = b'iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=='
```
Chatewgne, line 121: You could also import this variable from `geotrek/common/utils/testdata`

geotrek/trekking/tests/test_parsers.py
marcantoinedupre: Oops, I missed that. Removed.

geotrek/trekking/tests/test_parsers.py
marcantoinedupre: Yes, I guess I did not want to add another dependency on the `common` module. But hey, It's what the `common` module is for!
I updated my code to import the test data from `common`.

Chatewgne (approved):

PR #3333, opened 2022-11-24
Add geometry is not valid check

LePetitTim:


geotrek/core/tests/test_forms.py
```diff
@@ -72,6 +72,15 @@ def test_hidden_fields(self):
...
         self.assertIsInstance(form.fields['departure'].widget, HiddenInput)
 
+    def test_invalid_geom_path(self):
+        user = UserFactory()
+        # Just intersecting
+        form1 = PathForm(
+            user=user,
+            data={'geom': '{"geom": "LINESTRING(2.5 45.5, 2.5 45.5)", "snap": [null, null]}'}
+        )
+        self.assertFalse(form1.is_valid(), str(form1.errors))
```
marcantoinedupre, line 82: Is it intended to have `str(form1.errors)` as the message of the assertion?

marcantoinedupre (approved):
Nice :+1:

PR #3336, opened 2022-11-30
Fix required field display, add direction on lines (with setting)

LePetitTim:


LePetitTim:
Direction per blade:


Direction per lines:

Chatewgne (approved):

PR #3350, opened 2022-12-07
Update data model

babastienne:


Chatewgne (approved):

Chatewgne (approved):

submarcos:
split data model update from fixture fix. add test to ensure fixtures are ok

Makefile
```diff
@@ -47,18 +47,18 @@ css:
 
 %.pdf:
 	mkdir -p docs/data-model
-	postgresql_autodoc -h localhost -u geotrek -d geotrekdb -t dot -m "$*_.*" --password=geotrek -s "public"
+	postgresql_autodoc -l /usr/share/postgresql-autodoc -h localhost -u geotrek -d geotrekdb -t dot -m "$*_.*" --password=geotrek -s "public"
```
submarcos, line 50:
```suggestion
	postgresql_autodoc -l /usr/share/postgresql-autodoc -h localhost -u geotrek -d geotrekdb -t dot -m "$*_.*" --password=geotrek -s "public"
```
Why using ``-l /usr/share/postgresql-autodoc`` ?

submarcos: no need to specify on my instance as it's default shared data for postgresql-autodoc

babastienne: Ok, on ubuntu 22.04 ?

I installed autodoc especially for this purpose and by default it seems that the program was trying to reach another folder (as mentioned in this issue https://github.com/cbbrowne/autodoc/issues/26).
If this is only a bug on my computer I guess we should remove this commit then.

PR #3356, opened 2022-12-12
:sparkles: Add poi shapeparser

LePetitTim:


docs/install/import.rst
```diff
@@ -339,6 +339,17 @@ Treks are now imported into your own instance.
 Import other datas from a file
 ==============================
 
+You can add parsers in your custom `parsers.py` file (``/opt/geotrek-admin/var/conf/parsers.py``) which will allow you to
+import datas from files directly in your admin (superusers only).
```
Chatewgne, line 343:
```suggestion
import data from files directly in your admin (superusers only).
```

docs/install/import.rst
```diff
@@ -339,6 +339,17 @@ Treks are now imported into your own instance.
...
+For exemple, some parsers are not available by default but you can use them adding some lines in your parsers file :
```
Chatewgne, line 344:
```suggestion
For example, some parsers are not available by default but you can still use them by adding some lines in your parsers file :
```

geotrek/trekking/parsers.py
```diff
@@ -38,6 +39,49 @@ def filter_duration(self, src, val):
...
+    }
+    topology = Topology.objects.none()
+
+    def start(self):
+        super().start()
+        if settings.TREKKING_TOPOLOGY_ENABLED and not Path.objects.exists():
+            raise GlobalImportError(_("You need to add a network of paths before importing POIs"))
+
+    def filter_geom(self, src, val):
+        if val is None:
```
Chatewgne, line 64:
```suggestion
```
This is checked again right after

Chatewgne (changes requested):

PR #3348, opened 2022-12-07
feat: add missing attachements on SensitivityArea APIs

lpofredc:
PR to fix missing attachments in SensitiveAreas APIs.

Tests updated to check if attachments are in API responses.

PR #3358, opened 2022-12-13
Biodivsports public detail view

lpofredc:
Add a new public web view on sensitive areas.
URL provided in sensitive area api under "public_url" key.

Build with VueJS which will also be used to integrate Biodiv-sports widget generator https://biodivsports-widget.lpo-aura.org/

Already manage attachments (cf. PR #3348 )

MicrosoftTeams-image (1)


lpofredc:
It adds a public webpage (no auth required, cf. previous screenshot) for each published sensitive area that will describe sensitive areas (map, sensitive periods, descriptions, contacts, attachments, etc.)  ex: https://biodiv-sports.fr/public/sensitiveareas/1296.

That page url is provided in SensitiveArea apis under `public_url` key and can be used by any client. This is an expected special need in the BiodivSports instance

image

babastienne:
@lpofredc could you try not to use VueJS as well as FontAwesome ? We try to reduce the number of dependencies on this project and those (especially Vue) are important librairies. It doesn't seems right to use VueJS if not to replace all views with it in Geotrek and create a better front-end.

It does seems that the use of external dependencies to create public views is not required in order to achieve the feature you're trying to add.

I'm open to discuss this subject with you if needed.

lpofredc:
Ok, I removed vueJS dependency to use django templates native features. I agree to not add so much dependecies.

A better queryset with select related and prefetch will be possible after PR #3348 merge.

It was also to prepare biodivsport widget generator integration that also use VueJS and to avoid coding reactivity features by using pure javascript or jQuery.

submarcos:
I suggest to add TemplateRenderer directly on SensitivityArea APIV2 viewset. If you call ``/api/v2/sensitivearea/1```, you get the json,  /api/v2/sensitivearea/1.html, you got HTML version.

lpofredc:
Ok, i moved this template view into `api/v2` module with add a global map preview for root url (`/api/v2/sensitivearea.html`).

> I suggest to add [TemplateHTMLRenderer](https://www.django-rest-framework.org/topics/html-and-forms/#rendering-html) directly on [SensitivityArea APIV2 viewset](https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/api/v2/views/sensitivity.py#L13). If you call `/api/v2/sensitivearea/1`, you get the json, /api/v2/sensitivearea/1.html, you got HTML version. No need to add view or unauthenticated url in sensitivity app. Api is already unauthenticated and filter published data

lpofredc:
After discussion, will probably be replaced by a new dev using [GeotrekCE/Geotrek-rando-widget](https://github.com/GeotrekCE/Geotrek-rando-widget )

camillemonchicourt:
Voir l'article de présentation technique du widget - https://makina-corpus.com/application-web-mobile/creer-application-en-tant-que-composant-web-stencil

PR #3385, opened 2022-12-20
🐛 Recreate cache folders if don't exist

submarcos:
Fix particular case where migrations flush and recreate cache folders, then user follow changelog directives and manually trash var/cache/* folder

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,6 +23,10 @@ In preparation for HD Views developments (PR #3298)
 - Bump django-clearcache to 1.2.1
 - Add libvips to dependencies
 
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -23,6 +23,10 @@ In preparation for HD Views developments (PR #3298)
 - Bump django-clearcache to 1.2.1
 - Add libvips to dependencies
 
+**Bug fixes**
+
+- Recreate cache folders if missing.
```
camillemonchicourt, line 28: Should reference https://github.com/GeotrekCE/Geotrek-admin/issues/3384

PR #3399, opened 2023-01-05
:bug: Force simple geometrycollection on sites

LePetitTim:


Chatewgne (approved):

geotrek/outdoor/migrations/0042_flatten_geometrycollection.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('outdoor', '0041_auto_20221110_1128'),
+    ]
+
+    operations = [
+        migrations.RunPython(flatten_geometrycollection),
```
submarcos, line 20:
```suggestion
        migrations.RunPython(flatten_geometrycollection, migrations.RunPython.noop),
```

submarcos: make reverse migration working

geotrek/outdoor/tests/test_models.py
```diff
@@ -25,6 +25,35 @@ def test_published_children_by_lang(self):
         SiteFactory(name='child3', parent=parent, published_fr=True)
         self.assertQuerysetEqual(parent.published_children, ['<Site: child2>', '<Site: child3>'])
 
+    def test_validate_collection_geometrycollection(self):
+        site_simple = SiteFactory.create(name='site',
+                                         geom='GEOMETRYCOLLECTION(POINT(0 0), POLYGON((1 1, 2 2, 1 2, 1 1))))')
+        site_simple.refresh_from_db()
```
submarcos, line 31: Ithink it would be better to refresh_from_db in save method, just after super().save(). If we create an instance by hand, real value is directly available

PR #3408, opened 2023-01-16
:bug: Fix apidae parser should not update every time

LePetitTim:


camillemonchicourt:
What was the problem ?
And is it for all APIDAE parsers ?

LePetitTim:
> What was the problem ? And is it for all APIDAE parsers ?

It's mostly for apidae parser, but i can improve other parsers.
Imports with this parser was always saying that there are some update which was not the case. I improved it.

camillemonchicourt:
OK, thanks.
So it will update data in Geotrek (and their update_date), only if data have changed in Apidae. 
Nice.

PR #3416, opened 2023-01-19
:bug: Fix Loaddem command update other types of geometry

LePetitTim:


Chatewgne (approved):

marcantoinedupre (approved):
Looks good :+1:

geotrek/altimetry/management/commands/loaddem.py
```diff
@@ -109,7 +109,10 @@ def handle(self, *args, **options):
         if update_altimetry_paths:
             if verbose:
                 self.stdout.write('Updating 3d geometries.\n')
-            Path.objects.all().update(geom=F('geom'))
+            for model in apps.get_models():
+                if 'geom' in [field.name for field in model._meta.get_fields()]:
+                    model.objects.all().update(geom=F('geom'))
+
```
submarcos, line 115: I think it is a little more complicated.

in case of topology enabled:
  update Path geom=geom
  update only models having 3D geometry and not derivating from Topology ?

in other cases : 
  update only having 3D geometry

submarcos: this can be so long ...

LePetitTim: True, i will do that then

PR #3418, opened 2023-01-19
:bug: Fix parser translation field,  add modified  when new value is different from old value (for each language)

LePetitTim:
#3408

### DIFF
--- a/geotrek/trekking/tests/test_parsers.py
+++ b/geotrek/trekking/tests/test_parsers.py
@@ -664,6 +664,13 @@ class TestApidaeTrekParser(ApidaeTrekParser):
     selection_id = 654321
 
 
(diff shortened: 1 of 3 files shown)


geotrek/trekking/tests/test_parsers.py
```diff
@@ -664,6 +664,13 @@ class TestApidaeTrekParser(ApidaeTrekParser):
     selection_id = 654321
 
 
+class TestApidaeTrekSameValueDifferentTraductionParser(TestApidaeTrekParser):
```
camillemonchicourt, line 667: Traduction or Translation ?

submarcos:
I don't understand the title, what is the goal ?

PR #3428, opened 2023-01-25
Fix: APIDAE Trek parser now raises an error on not continuous segments

marcantoinedupre:
If the `merge` action from django/gis's GEOS object fails to produce a single LineString feature a RowImportError exception is raised and the Trek from APIDAE is not imported.

This is necessary because the altimetry profile cannot be computed for geometries which are not a continuous LineString.

### DIFF
--- a/geotrek/trekking/parsers.py
+++ b/geotrek/trekking/parsers.py
@@ -962,6 +962,8 @@ class ApidaeTrekParser(AttachmentParserMixin, ApidaeBaseTrekkingParser):
         geos = ApidaeTrekParser._convert_to_geos(first_feature.geom)
         if geos.geom_type == 'MultiLineString':
             geos = geos.merged
+            if geos.geom_type == 'MultiLineString':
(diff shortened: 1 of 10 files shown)


geotrek/trekking/parsers.py
```diff
@@ -957,6 +957,8 @@ def _maybe_get_linestring_from_layer(layer):
         geos = ApidaeTrekParser._convert_to_geos(first_feature.geom)
         if geos.geom_type == 'MultiLineString':
             geos = geos.merged
+            if geos.geom_type == 'MultiLineString':
+                raise RowImportError("The geometry cannot be converted to a single continuous LineString feature")
```
numahell, line 961: This may be translated here ? (not sure if error messages are translated)

marcantoinedupre: `RowImportErrors` show up in the import report so they should indeed be translated. Done.

Chatewgne (approved):

Issue #3494, opened 2023-02-22
SensitiveAreas details page does not show nearby Sites/Courses (and the other way around)


marcantoinedupre:
To reproduce : create a sensitive area overlapping a site or course and visit the detail page of the sensitive area. The site or course does not appear in the list of nearby objects. (visit the site or course detail page, the sensitive area does not appear in the list either).

![image](https://user-images.githubusercontent.com/15977920/220605430-1605aa6e-4088-4846-9a2c-8f0b83abc78a.png)

PR #3495, opened 2023-02-22
Fix sensitive areas nearby outdoor objects (issue #3494)

marcantoinedupre:
issue #3494

### DIFF
--- a/geotrek/sensitivity/models.py
+++ b/geotrek/sensitivity/models.py
@@ -296,6 +296,41 @@ if 'geotrek.tourism' in settings.INSTALLED_APPS:
                                lambda self: intersecting(tourism_models.TouristicEvent, self, 0).filter(published=True),
                                _("Published touristic events"))
 
+
+if 'geotrek.outdoor' in settings.INSTALLED_APPS:
+    from geotrek.outdoor import models as outdoor_models
+
+    outdoor_models.Site.add_property('sensitive_areas',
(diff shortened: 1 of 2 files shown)


geotrek/sensitivity/models.py
```diff
@@ -296,6 +300,41 @@ def pretty_practices(self):
...
+                               _("Touristic contents"))
+    SensitiveArea.add_property('published_sites',
+                               lambda self: intersecting(outdoor_models.Site, self).filter(published=True),
+                               _("Published touristic contents"))
+    SensitiveArea.add_property('courses',
+                               lambda self: intersecting(outdoor_models.Course, self),
+                               _("Touristic events"))
+    SensitiveArea.add_property('published_courses',
+                               lambda self: intersecting(outdoor_models.Course, self).filter(published=True),
+                               _("Published touristic events"))
```
submarcos, line 335: you should add outdoor related properties in outdoor models.py

marcantoinedupre: Do you mean to replace lambdas with actual methods on outdoor models?

marcantoinedupre: We have discussed the matter. The `sensitive_areas` properties cannot be added to Site and Course in the `geotrek.outdoor.models` module because of a circular import (`geotrek.sensitivity.models` already imports the outdoor module so the outdoor module cannot import sensitivity to get the `SensitiveArea` class).

PR #3512, opened 2023-03-10
Perf improve restrictedarea filter

LePetitTim:


geotrek/maintenance/filters.py
```diff
@@ -23,18 +24,17 @@ def get_geom(self, value):
...
-            for element in qs:
-                if element.target:
-                    if not element.target.geom or element.target.geom.intersects(bbox):
-                        interventions.append(element.pk)
-                elif not element.target and element.target_type:
+        for element in qs:
+            if element.target:
+                if not element.target.geom or element.target.geom.intersects(bbox):
                     interventions.append(element.pk)
+            elif not element.target and element.target_type:
```
submarcos, line 36:
```suggestion
            elif element.target_type:
```

geotrek/maintenance/filters.py
```diff
@@ -23,18 +24,17 @@ def get_geom(self, value):
     def filter(self, qs, values):
         if not values:
             return qs
-        if not isinstance(values, list):
-            values = [values]
+        if isinstance(values, list):
+            bbox = GeometryCollection([self.get_geom(value) for value in values])
```
submarcos, line 28:
```suggestion
            bbox = GeometryCollection([self.get_geom(value) for value in values])
```
I think it's not a bbox in this case

submarcos (approved):

Chatewgne (approved):

Issue #3321, opened 2022-11-15
The API should convert relative media path to absolute 


dtrucs:
From the description fields, it is possible to attach media. The `src` attribute is filled with a relative path instead of an absolute path.  
Whatever service connects to the API should not parse the content to add the base path of the admin URL.

For example: 
A description given by a trek endpoint https://admin-pyrenees-orientales.fr/api/v2/trek/254789/?fields=description
And below the UI for the user ⬇️ 
![image](https://user-images.githubusercontent.com/1926041/201939755-3b383ccf-2851-47cd-96fa-e0a5e214dd28.png)

Chatewgne:
The relative path in question when displaying the source code for this description :

![image](https://user-images.githubusercontent.com/26015573/201954826-c601fcb0-8271-42f5-8a0d-94049e8e9cad.png)

- [ ] This pictogram insertion tool should insert an absolute URL in the `src` tag
- [ ] It would be useful to write a migration for the old data, converting fields with relative paths to absolute paths, or at least a command that helps finding all fields that need an update

camillemonchicourt:
Hum, what about if we change the Geotrek-admin URL.
Not sure storing an absolute path is the good solution...

dtrucs:
> Hum, what about if we change the Geotrek-admin URL. Not sure storing an absolute path is the good solution...

Totally agree 👍  
Ideally, I think the conversion should be done by the API looking at its basePath. However parsing the content with a regexp could go wrong (false positive).

Anyway, it is not flexible to work with stored HTML. Maybe in the future storing the content in markdown would make the manipulation more flexible.

bastyen:
Geotrek mobile handle this case here : https://github.com/GeotrekCE/Geotrek-mobile/blob/main/src/app/services/settings/settings.service.ts#L290

camillemonchicourt:
Fait dans la 2.95.0, mais je ne vois pas le résultat quand j'affiche une description avec un service intégré dans celle-ci : 

![image](https://user-images.githubusercontent.com/4418840/215331656-694e9c77-9999-4e85-8b16-b686d4cd156d.png)

Le chemin du picto du service est relatif : ``<img src=\"/media/upload/descent-icon.svg\" alt=\"Descente\" />``

camillemonchicourt:
Corrigé dans la 2.98.1

PR #3579, opened 2023-04-25
Fix replacement of service pictogram paths with absolute URLs in API

marcantoinedupre:
Also expand to 'description_teaser' and 'ambiance' fields.

Should address issue [#3321](https://github.com/GeotrekCE/Geotrek-admin/issues/3321)

### DIFF
--- a/geotrek/api/v2/serializers.py
+++ b/geotrek/api/v2/serializers.py
@@ -775,8 +775,7 @@ if 'geotrek.trekking' in settings.INSTALLED_APPS:
                 imgs = soup.find_all('img')
                 for img in imgs:
                     if img.attrs['src'][0] == '/':
-                        base_url = self.context.get("request").build_absolute_uri("/")[:-1]
-                        img['src'] = base_url + img.attrs["src"]
+                        img['src'] = self.context.get("request").build_absolute_uri(img.attrs["src"])
                 # Note: beautifulsoup makes a valid HTML doc from the data,
                 # but we don't want <html> or <body> in the output
                 return "".join([str(c) for c in soup.find("body").children])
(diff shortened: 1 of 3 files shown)


geotrek/api/v2/serializers.py
```diff
@@ -788,6 +767,28 @@ def get_departure_city(self, obj):
...
 
+        def _replace_image_paths_with_urls(self, data):
+            def replace(html_content):
+                if not html_content:
+                    return html_content
+                soup = BeautifulSoup(html_content, features="lxml")
+                imgs = soup.find_all('img')
+                for img in imgs:
+                    if img.attrs['src'][0] == '/':
+                        base_url = self.context.get("request").build_absolute_uri("/")[:-1]
```
submarcos, line 778: try to use directly build_absolute_url to get final url : self.context.get("request").build_absolute_uri(img.attrs['src'])

geotrek/api/v2/serializers.py
```diff
@@ -788,6 +767,28 @@ def get_departure_city(self, obj):
...
+                        img['src'] = base_url + img.attrs["src"]
+                # Note: beautifulsoup makes a valid HTML doc from the data,
+                # but we don't want <html> or <body> in the output
+                return "".join([str(c) for c in soup.find("body").children])
```
submarcos, line 782: I think there is an option or method in beautifulsoup for this

geotrek/api/v2/serializers.py
marcantoinedupre: Right, I have done it.

geotrek/api/v2/serializers.py
marcantoinedupre: Right my complicated statement is useless. By passing `features="html.parser"` to BeautifulSoup constructor no extra HTML tags are added to the input (no <html> or <body>).

PR #3610, opened 2023-06-07
✨ Use practice color of treks used for mobile in pdfs

LePetitTim:


geotrek/trekking/models.py
```diff
@@ -68,7 +68,7 @@ class Practice(TimeStampedModelMixin, PictogramMixin):
     order = models.IntegerField(verbose_name=_("Order"), null=True, blank=True,
                                 help_text=_("Alphabetical order if blank"))
     color = ColorField(verbose_name=_("Color"), default='#444444',
-                       help_text=_("Color of the practice, only used in mobile."))  # To be implemented in Geotrek-rando
+                       help_text=_("Color of the practice, only used in mobile and pdfs."))  # To be implemented in Geotrek-rando
```
submarcos, line 71: if you change a tanslated text, you should update and fix translation files

Issue #3458, opened 2023-01-31
[API] Add an array filter of selected identifiers for endpoints of type `list`.


dtrucs:
It would probably be relevant to be able to add a filter for API lists like `/post/?id[]=1,2,3` so that only certain items are selected as a list.
Probably the term `id[]` is not the most appropriate, but that's to get the idea.

I can see an use case with the `children` field from `outdoor_site` endpoint that would avoid as many API calls as children item in frontend part like:
```js
const children = await Promise.all([1,2,3].map(id => fetch(`api.com/api/v2/outdoor_site/${id}`)
```
 (cf https://github.com/GeotrekCE/Geotrek-rando-v3/issues/846).

NB: I am thinking of the API cache and the consequences when there is a mutation of an ID child for example.

dtrucs:
This feature could be also useful to optimize fetching data for the GTR homepage suggestions.

dtrucs:
Thank for the contribution

camillemonchicourt:
We usually close Issues when the related PR is merged and available in an official release.

PR #3620, opened 2023-06-23
Enable filtering lists by objects IDs on APIv2

Chatewgne:
https://github.com/GeotrekCE/Geotrek-admin/issues/3458


marcantoinedupre (approved):
:+1:

PR #3626, opened 2023-06-29
Fix: protect report endpoint from email failure

marcantoinedupre:
Previously when the confirmation email sending failed the endpoint answered with error 500 even though the report was correctly saved.


docs/changelog.rst
```diff
@@ -733,12 +734,8 @@ Do not use - Causes bug in Geotrek-Rando-v3 exposing Outdoor data
 - Add a scroll bar into filter form and module list (#2849)
 - In projects, start year must be before end year (#3567)
 
-**Maintenance**
-
-- Upgrade `django-mapentity`
-
 
-2.98.1     (2023-05-30)
+2.98.1 (2023-05-30)
```
submarcos, line 738:
```suggestion
2.98.1     (2023-05-30)
```

submarcos:
https://docs.djangoproject.com/en/5.1/topics/email/#django.core.mail.send_mail

send_mail can be used with fail_silently=True

PR #3643, opened 2023-07-18
Filter trek labels according to whether they are published or not

Huquette:
We want to be able to filter the labels on treks according to whether the labels are published or not.

#3640


marcantoinedupre (approved):
Nice :+1:

numahell (changes requested):
Nice PR, just a little suggestion :)

submarcos (commented):
In other models, logic is to define published=False [by default](https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/trekking/templates/trekking/sql/post_90_defaults.sql#L84)

Maybe it is specific to labels to keep True by default

Chatewgne:
> In other models, logic is to define published=False [by default](https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/trekking/templates/trekking/sql/post_90_defaults.sql#L84)
> 
> Maybe it is specific to labels to keep True by default

Maybe we should use "default=False" but use the migration to publish pre-existing labels so that it doesn't affect existing instances ?

marcantoinedupre:
Either way seems fine to me. Though I have just thought that maybe `default=True` could spare us some troubles with imports from third-parties SIT.

Chatewgne (approved):

PR #3724, opened 2023-09-13
fix: cities display on string  (refs #3585)

juggler31:
Avant: La liste des "cities" était retournée dans un tableau.

Maintenant: Elle est retournée sous forme de "string" avec chaque ville / commune séparées par des virgules

### DIFF
--- a/geotrek/infrastructure/models.py
+++ b/geotrek/infrastructure/models.py
@@ -148,7 +148,13 @@ class BaseInfrastructure(BasePublishableMixin, Topology, StructureRelated):
 
     @property
     def cities_display(self):
-        return [str(c) for c in self.cities] if hasattr(self, 'cities') else []
+        all_cities_name = []
+        if hasattr(self, 'cities'):
+            for city in self.cities:
+                all_cities_name.append(city.name)
+            return ", ".join(all_cities_name)
+        else:
+            return ""
 
     @classproperty
     def cities_verbose_name(cls):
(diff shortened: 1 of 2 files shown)


geotrek/infrastructure/tests/test_views.py
```diff
@@ -52,7 +52,7 @@ def get_expected_json_attrs(self):
 
     def get_expected_datatables_attrs(self):
         return {
-            'cities': '[]',
+            'cities': '',
```
Chatewgne, line 55: Maybe add some cities around the infrastructure geom to test list result ? See `CityFactory`

Chatewgne (changes requested):

babastienne (changes requested):
- No tests covering the new code
- Missing a changelog entry

submarcos:
Maybe you should just override datatable serializer to provide right formatted values. Model property is probably used elsewhere

geotrek/infrastructure/models.py
```diff
@@ -148,7 +148,13 @@ def type_display(self):
 
     @property
     def cities_display(self):
-        return [str(c) for c in self.cities] if hasattr(self, 'cities') else []
+        all_cities_name = []
+        if hasattr(self, 'cities'):
```
Chatewgne, line 152: With Django you could do :
```suggestion
        if hasattr(self, 'cities'):
            all_cities_name = list(self.cities.objects.values_list('name', flat=True))
```
It think it's better to use SQL queryset to retrieve only the name from db

Chatewgne (changes requested):

geotrek/infrastructure/tests/test_views.py
Chatewgne: Need to add tests to cover lines requested by codecov

PR #3755, opened 2023-09-27
Fix migration check in CI

submarcos:
## Description

By using ENV=tests, CI doesn't run migrations and can't check if migrations are missing in code


marcantoinedupre (approved):
Good point!

Why had we deactivated `makemigrations` in test environments in the first place?

submarcos:
> `makemigrations`

I'm not sure but this : https://github.com/GeotrekCE/Geotrek-admin/blob/master/geotrek/settings/env_tests.py#L54

disable makemigrations / module en test env. I don't known why migrations are disabled whereas they installed sql triggers that are tested in CI...

PR #3968, opened 2024-02-29
✨ [FEAT] Add CirkwiParser to retrieve Treks and Touristic Contents (refs #3947)

Chatewgne:
## Related Issue

https://github.com/GeotrekCE/Geotrek-admin/issues/3947

## Checklist

- [x] I have followed the guidelines in our [Contributing document](https://geotrek.readthedocs.io/en/latest/CONTRIBUTING.html)
- [x] My code respects the Definition of done available in the [Development section of the documentation](https://geotrek.readthedocs.io/en/latest/contribute/development.html#definition-of-done-for-new-model-fields)
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes
- [x] I added an entry in the changelog file
- [x] My commits are all using prefix convention (emoji + tag name) and references associated issues
- [x] I added a label to the PR corresponding to the perimeter of my contribution
- [x] The title of my PR mentionned the issue associated

### DIFF
--- a/geotrek/cirkwi/parsers.py
+++ b/geotrek/cirkwi/parsers.py
@@ -5,12 +5,12 @@ from django.conf import settings
 from django.contrib.gis.geos import Point, MultiPoint, GEOSGeometry
 from django.utils.translation import gettext as _
 
+from geotrek.common.utils.parsers import get_geom_from_gpx
 from geotrek.trekking.models import DifficultyLevel
 from geotrek.cirkwi.models import CirkwiLocomotion
 from geotrek.common.parsers import AttachmentParserMixin, GlobalImportError, Parser, RowImportError
 from geotrek.tourism.models import TouristicContent, TouristicContentType1
 from geotrek.trekking.models import Trek, Practice
-from geotrek.trekking.parsers import ApidaeTrekParser
 
 
 class CirkwiParser(AttachmentParserMixin, Parser):
@@ -154,7 +154,7 @@ class CirkwiTrekParser(CirkwiParser):
 
     def filter_geom(self, src, val):
         response = self.request_or_retry(url=val)
-        return ApidaeTrekParser._get_geom_from_gpx(response.content)
+        return get_geom_from_gpx(response.content)
 
     def filter_practice(self, src, val):
         """
--- a/geotrek/cirkwi/parsers.py
+++ b/geotrek/cirkwi/parsers.py
@@ -1,0 +1,310 @@
+import requests
+import xml.etree.ElementTree as ET
+
+from django.conf import settings
+from django.contrib.gis.geos import Point, MultiPoint, GEOSGeometry
+from django.utils.translation import gettext as _
+
+from geotrek.trekking.models import DifficultyLevel
+from geotrek.cirkwi.models import CirkwiLocomotion
+from geotrek.common.parsers import AttachmentParserMixin, GlobalImportError, Parser, RowImportError
+from geotrek.tourism.models import TouristicContent, TouristicContentType1
+from geotrek.trekking.models import Trek, Practice
+from geotrek.trekking.parsers import ApidaeTrekParser
+
+
+class CirkwiParser(AttachmentParserMixin, Parser):
+    """
+    auth: Allows to configure HTTP auth on parser requests
+    create: Create a Cirkwi Locomotion from label, before trying to map it to a Trek Practice with same label
+    delete: Delete old objects that are now missing from flux (based on 'get_to_delete_kwargs')
+    default_language: Allows to define which language this parser will populate by default
+    eid: Field to use as Cirwki id
+    provider: Allows to differentiate multiple Parser for the same model
+    rows: Quantity of objects to load per page (Cirkwi pagination)
+    update_only: Do not delete previous objects, and query Cirkwi API with most recent date_update
+    """
+    auth = ()
+    create = False
+    delete = False
+    default_language = settings.MODELTRANSLATION_DEFAULT_LANGUAGE
+    eid = 'eid'
+    provider = "Cirkwi"
+    rows = 10
+    update_only = False
+
+    field_options = {
+        "geom": {"required": True},
+        "name": {"required": True},
+    }
+    constant_fields = {
+        'published': True,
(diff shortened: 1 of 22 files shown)


geotrek/cirkwi/parsers.py
```diff
@@ -0,0 +1,310 @@
...
+                             "infos_parcours/info_parcours/adresse/position/lng/*"),
+        "geom": "fichier_trace@@url",
+        "practice": ("locomotions/locomotion@@type", "locomotions/locomotion@@id_locomotion"),
+        "difficulty": "locomotions/locomotion@@difficulte",
+        "duration": "locomotions/locomotion@@duree",
+    }
+
+    def filter_geom(self, src, val):
+        response = self.request_or_retry(url=val)
+        return ApidaeTrekParser._get_geom_from_gpx(response.content)
```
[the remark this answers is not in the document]
marcantoinedupre: I agree we should refactor this. To use a protected static method from an unrelated class is not good. Regarding the GPX part from the `ApidaeTrekParser` I suggest to transform the code into functions in a `utils` module. For instance `geotrek.common.utils.parsers`.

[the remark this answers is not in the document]
Chatewgne: Thanks for the feedback :)

geotrek/trekking/parsers.py
```diff
@@ -690,13 +690,16 @@ def filter_geom(self, src, val):
...
-            return ApidaeTrekParser._get_geom_from_kml(kml_file)
+        try:
+            if ext == 'gpx':
+                return get_geom_from_gpx(geom_file)
+            elif ext == 'kml':
+                return get_geom_from_kml(geom_file)
+            elif ext == 'kmz':
+                kml_file = zipfile.ZipFile(io.BytesIO(geom_file)).read('doc.kml')
+                return get_geom_from_kml(kml_file)
+        except ValueError as e:
```
marcantoinedupre, line 701: The messages of errors raised by the get geom functions are user-readable and intended to be displayed in the parser import output. So raising the error again as a `RowImportError` with the original message is good. :+1: 
IMO `ValueError` is too broad for this usecase. The `try` clause is above too much code to use it. The risk is to catch `ValueError` exceptions from unexpected part of the code and to show them as is to the user which is not great. And developers will struggle to find the origin of such errors because they won't have a traceback in the import output.

I suggest to add a new type

```python
class GeomValueError(Exception):
  pass
```

in the `utils` module and to `try..catch` that instead.

marcantoinedupre (changes requested):
I have reviewed the move of the get_geom* functions and I have added a comment. :)

geotrek/trekking/parsers.py
Chatewgne: Good point, thank you !

geotrek/cirkwi/parsers.py
```diff
@@ -0,0 +1,302 @@
...
+        # Extract XML attributes
+        elif src.startswith('@@'):
+            return val.attrib[src[2:]]
+        else:
+            # Replace language attribute
+            if "'<LANG>'" in src:
+                src = src.replace("<LANG>", self.default_language)
+            # Return a list of XML elements
+            if src.endswith('/*'):
+                return val.findall(src[:-2])
```
marcantoinedupre, line 110: I think it is possible to return directly strings here instead of a list of XML elements. This way it would work in the same way than when a single text value is returned at line 114. And you could remove all conversion done in `filter_*` methods.

geotrek/cirkwi/parsers.py
```diff
@@ -0,0 +1,302 @@
...
+from geotrek.common.utils.parsers import get_geom_from_gpx
+from geotrek.trekking.models import DifficultyLevel
+from geotrek.cirkwi.models import CirkwiLocomotion
+from geotrek.common.parsers import AttachmentParserMixin, Parser, RowImportError
+from geotrek.tourism.models import TouristicContent, TouristicContentType1
+from geotrek.trekking.models import Trek, Practice
+
+
+class CirkwiParser(AttachmentParserMixin, Parser):
+    """
```
marcantoinedupre, line 16: I suggest to expand the docstring with explanations regarding the new XML-related possibilities for the fields mapping. Here is a proposal :

```
The CirkwiParser class handles some XML-related specificities for the fields mapping:

- `src` values should follow an XML path syntax so the separator is '/' instead of '.'
- `src` value "locomotions/locomotion" gets the text value of the first <locomotion> XML element encountered (inside a <locomotions> element at the row's root level)
- "informations/information[@langue='fr']/titre" means we get the text value of the first <titre> element having a parent <information> with attribute `langue` equals "fr"
- the "<LANG>" marker is replace with the `default_language` of the parser
- a '@@' sequence at the end followed by a attribute name indicates the fetched value should be the value of the attribute (for instance "locomotions/locomotion@@type" will get the value of the 'type' attribute of the <locomotion> element)
- finally the '/*' sequence at the end of a `src` value indicates a list of all XML elements matching the path should be returned (so "locomotions/locomotion/*" has the same meaning than "locomotions.*.locomotion" from the base Parser class)

See the overridden `CirkwiParser.get_part` method for the implementation.
```

Probably too verbose feel free to include it or not as you like.

marcantoinedupre (approved):
I have made a few remarks you may address.

LGTM otherwise. Nice work! :+1:

geotrek/cirkwi/parsers.py
Chatewgne: It gets tricky with lists here, because, what do we do after retrieving the multiple elements ? Do we need to get the list of inner texts, or a list from the attributes ? Are we looking for one or several other inner element(s) ? We would need to add some more recursivity to combine this with the other syntaxes. It's why I chose to use `get_part` to return the list of elements and let `filter_xxxx` handle further processing for now

marcantoinedupre: Alright, let's leave it as is then.

PR #3979, opened 2024-03-04
Prepare Django 4.2 compatibility

submarcos:
Prepare Django 4.2 compatibility without dependency update.

Django 4.2 will break compatibility with postgreSQL < 12

This PR made just code compatible and ready to update


geotrek/tourism/locale/de/LC_MESSAGES/django.po
```diff
@@ -97,7 +97,7 @@ msgid "Phone"
 msgstr "Telefon-Nummer"
 
 msgid "Email"
-msgstr "Email"
+msgstr "Courriel"
```
submarcos, line 100:
```suggestion
msgstr ""
```

PR #4329, opened 2024-10-10
Fix 'loadinfrastructure' condition adding to infra without 'eid-field' option

noelmartinon:
If '--confition-field' is used with 'loadinfrastructure', add condition to infra even if option '--eid-field' is not used.
GeotrekCE#4328


Chatewgne (changes requested):
Thank you for contributing, could you please add a line at the top of `docs/changelog.rst` mentioning the fix and the issue number please ?

camillemonchicourt:
La PR du changelog de cette modification a été mergée (https://github.com/GeotrekCE/Geotrek-admin/pull/4329), mais pas celle-ci qui comprend la correction de la commande.

PR #4369, opened 2024-11-15
🐛 [BUG] Fix SchemaRandonneeParser url update (refs #4022)

justinefricou:
## Description

- Fix: a `Trek`'s URL being duplicated during a new import when the new `Trek`'s description was `null` or the same as the old description
- Mod : description and url are now parsed in a `filter_description` method rather than in the `end` method
- #4022 

## Checklist

- [x] I have followed the guidelines in our [Contributing document](https://geotrek.readthedocs.io/en/latest/contribute/contributing.html)
- [x] My code respects the Definition of done available in the [Development section of the documentation](https://geotrek.readthedocs.io/en/latest/contribute/development.html#definition-of-done-for-new-model-fields)
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes
- [x] I added an entry in the changelog file
- [x] My commits are all using prefix convention (emoji + tag name) and references associated issues
- [x] I added a label to the PR corresponding to the perimeter of my contribution
- [x] The title of my PR mentionned the issue associated

### DIFF
--- a/geotrek/trekking/parsers.py
+++ b/geotrek/trekking/parsers.py
@@ -1338,7 +1338,7 @@ class SchemaRandonneeParser(AttachmentParserMixin, Parser):
         'arrival': 'arrivee',
         'duration': 'duree',
         'difficulty': 'difficulte',
-        'description': 'instructions',
+        'description': ('instructions', 'url'),
         'ambiance': 'presentation',
         'description_teaser': 'presentation_courte',
         'advice': 'recommandations',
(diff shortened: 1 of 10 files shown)


geotrek/trekking/parsers.py
```diff
@@ -1588,3 +1592,11 @@ def end(self):
                 otc.save()
                 order += 1
         super().end()
+
+    @staticmethod
+    def find_url_in_description(trek):
```
Chatewgne, line 1597: Since this method is only used in `test_parsers.py`, can you move it to the class `SchemaRandonneeParserTests`  please ? That way it also won't need to be static

Chatewgne (changes requested):

geotrek/trekking/parsers.py
justinefricou: The name of the method might need to be modified so it's understandable that not any url will be retrieved.

Maybe `find_source_file_url_in_description` since this is how this url is described in `schema_randonnee`?
https://github.com/PnX-SI/schema_randonnee/blob/9a7e4a826e9af22c0a83b718a67a24b476c0c807/schema.json#L87

Chatewgne (approved):

